PR #3, opened 2015-03-10
Added support for logging nested properties of session items.

Giorgi:


304NotModified:
This is a breaking change?

Giorgi:
I guess it is. In the current version it will cause an exception and consequently won't log anything in the log. With this change it will log whatever the nested property refers too so I guess it does not do any harm.

304NotModified:
Well hou can have session items with a dot? I think it should be backwards compatible.  Maybe an extra boolean?

Also please post an example (config, usage)  thanks!

304NotModified:
Also unit test is also needed.

Giorgi:
Yes, you are right, you can use dotted names and this is going to break it. Do you think it's better to add a new bool property or create a separate `AspNetSessionNestedValueLayoutRenderer` or `something like that?

304NotModified:
I prefer the first solution because the code/functionality is highly related.

Giorgi:
OK, I will work on it. Any suggestions about the boolean property name?

304NotModified:
`isNestedVariable`?

Giorgi:
What about evaluateNestedProperties?

304NotModified:
EvaluateAsNestedProperties?

Giorgi:
Sounds good.

Giorgi:
@304NotModified Pull Request updated.

304NotModified:
Thanks! 

I will merge this after some unit tests and example configs.

Giorgi:
To be honest I am not sure how to mock Session. As for example configs do you want me to add it to the main Nlog project or this one?

304NotModified:
I would like a XML config example in this PR thread. For the wiki and other possible the news post.

304NotModified:
About mocking `session`, I agree it's with plain .Net difficult. (or just a hell).

I propose to use [Moq](https://www.nuget.org/packages/Moq/) on the unit test project.

304NotModified:
@Xharze where are the unit tests located for `Nlog.web`? It should be in the repository?

Giorgi:
@304NotModified Here is a sample xml configuration.

To have the nested properties of session object evaluated use the following layout for AspNetSession layout renderer

    <?xml version="1.0" encoding="utf-8" ?>
    <nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">

        <targets>
            <target name="logfile" xsi:type="File" fileName="file.txt" />
        </targets>

        <rules>
            <logger name="*" minlevel="Info" writeTo="logfile" layout="${aspnet-session:Variable=Username:EvaluateAsNestedProperties=true} ${date} ${message}/>
        </rules>
    </nlog>

Where do I write tests for it?

304NotModified:
@Xharze we don;t have unit tests in this repos available ?

Giorgi:
@304NotModified  Can you merge it without tests? I don't see any tests for existing code either.

304NotModified:
note this has been released and documented: https://github.com/NLog/NLog/wiki/AspNetSession-layout-renderer

PR #9, opened 2015-09-01
Added RequestContext + renderer

304NotModified:
A request context and renderer. The added variables are only valid for one request and is implemented by the `HttpContext.Items`. There will be no data leaks between requests with this approach.

Add variables to the request context:

```c#
RequestContext.Current["a"] = "b"
RequestContext.Current["numer"] = 2;
RequestContext.Current["d2"] = new DateTime(2020,1,30);;
```

to render, which results always in a `string`:

```xml
${aspnet-request-context:a} //"b"
${aspnet-request-context:number} //"2"
${aspnet-request-context:d2:format=yyyy-MM-dd} //"2020-01-30"
```

PR #11, opened 2015-10-14
Implement aspnet-item renderer

dafanasiev:
Implement aspnet-item renderer: render  value with specified name from HttpContext.Current.Items dictionary.

### DIFF
--- a/NLog.Web/LayoutRenderers/AspNetItemValueLayoutRenderer.cs
+++ b/NLog.Web/LayoutRenderers/AspNetItemValueLayoutRenderer.cs
@@ -1,0 +1,92 @@
+﻿using System;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+using System.Web;
+using NLog.Config;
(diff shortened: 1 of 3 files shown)


304NotModified:
Thanks!

Just looked in short to this:

1. This is almost the same as https://github.com/NLog/NLog.Web/pull/9, expect #9 prefixes the values. Do you think we should merge this?
2. I see some code duplication with [`AspNetSessionValueLayoutRenderer ` (EvaluateAsNestedProperties)](https://github.com/NLog/NLog.Web/blob/6a6991fa23da2dffc09a60afc013ad9db57b36cf/NLog.Web/LayoutRenderers/AspNetSessionValueLayoutRenderer.cs#L78-L89). Can you fix that?

304NotModified:
Bump

304NotModified:
Thanks for the init PR. I like this feature and so I completed the last things. Will release this in NLog.web 4.1

PR #14, opened 2015-10-25
Update license

mmaitre314:
Use BSD from https://raw.githubusercontent.com/NLog/NLog/master/LICENSE.txt and update copyright


304NotModified:
Thanks @mmaitre314 

Sorry for the late reply. The problem is here that I'm not the (only) official owner and it's very difficult to contact the real owners. That's also the reason why this license stuff isn't fixed yet :(

dabide:
So does this mean that the NLog.Web is licensed under the GPL?

304NotModified:
No it's BSD. Discussed that in the past with @Xharze.  But I think @Xharze is the only one who can legally change the headers.

304NotModified:
Superseded by #48

Issue #20, opened 2016-01-07
Consider creating an AspNetLayoutRendererBase class


epignosisx:
I just went through the exercise of creating a few custom Layout Renderers for AspNet and I experienced some pain points that I wanted to share and provide a solution.

Problem
================

1- Looking at the LayoutRenderers that are part of the NLog.Web package there is a bit of duplication in all of them in the form:

```c#
    HttpContext context = HttpContext.Current;
    if (context == null)
    {
        return;
    }

    var httpRequest = context.Request;
    if (httpRequest == null)
    {
        return;
    }
```

2- The raw usage of the static **HttpContext.Current** in the layout renderers makes it hard to unit test. I looked at how the NLog.Web layout renderers were unit tested and most do not have unit tests. Maybe this is part of the reason.

Proposal
=================

Introduce the following two concepts: **IHttpContextProvider** and **AspNetLayoutRendererBase**:

```c#
    public interface IHttpContextProvider
    {
        HttpContextBase Get(); //returns HttpContextBase which is unit testable.
    }

    public class DefaultHttpContextProvider : IHttpContextProvider
    {
        public HttpContextBase Get()
        {
            if (HttpContext.Current == null)
                return null;
            return new HttpContextWrapper(HttpContext.Current);
        }
    }

    public abstract class AspNetLayoutRendererBase : LayoutRenderer
    {
        private readonly IHttpContextProvider _httpContextProvider;

        //this is the constructor that would run during "normal" mode
        protected AspNetLayoutRendererBase() : this(new DefaultHttpContextProvider())
        {
        }

        //this constructor can be used during unit tests to pass an implementation will return a
        //a mocked HttpContextBase
        protected AspNetLayoutRendererBase(IHttpContextProvider httpContextProvider)
        {
            _httpContextProvider = httpContextProvider;
        }

        //Expose a property that subclasses can use instead of the static HttpContext.Current
        public HttpContextBase HttpContext
        {
            get { return _httpContextProvider.Get(); }
        }

        protected override void Append(StringBuilder builder, LogEventInfo logEvent)
        {
            //do this check in the base class instead of in each Layout Renderer
            if (HttpContext == null)
                return;

            //subclasses implement this method with the assurance that they will have a valid HttpContext.
            AppendCore(builder, logEvent);
        }

        protected abstract void AppendCore(StringBuilder builder, LogEventInfo logEvent);
    }
```

Here is a sample Layout Renderer:

```c#
[...]

PR #22, opened 2016-01-12
Introducing ASP.NET Layout Renderer base class

epignosisx:
This is an initial commit for #20. There is still more work to be done. I just wanted to share and get some initial feedback.

- [x] Introduce AspNetLayoutRendererBase
- [x] Introduce IHttpContextAccessor
- [x] Move existing layout renderers to use new base class
- [x] Add unit tests to existing layout renderers now that mocking HttpContext is easier.

fixes https://github.com/NLog/NLog.Web/issues/20

### DIFF
--- a/NLog.Web/LayoutRenderers/AspNetItemValueLayoutRenderer.cs
+++ b/NLog.Web/LayoutRenderers/AspNetItemValueLayoutRenderer.cs
@@ -36,7 +36,7 @@ namespace NLog.Web.LayoutRenderers
     /// </code>
     /// </example>
     [LayoutRenderer("aspnet-item")]
-    public class AspNetItemValueLayoutRenderer : LayoutRenderer
+    public class AspNetItemValueLayoutRenderer : AspNetLayoutRendererBase
     {
         /// <summary>
         /// Gets or sets the item variable name.
@@ -56,18 +56,14 @@ namespace NLog.Web.LayoutRenderers
         /// </summary>
         /// <param name="builder">The <see cref="StringBuilder"/> to append the rendered data to.</param>
         /// <param name="logEvent">Logging event.</param>
-        protected override void Append(StringBuilder builder, LogEventInfo logEvent)
+        protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
         {
             if (Variable == null)
             {
                 return;
             }
 
-            HttpContext context = HttpContext.Current;
-            if (context == null)
-            {
-                return;
-            }
+            HttpContextBase context = HttpContextAccessor.HttpContext;
 
             var value = PropertyReader.GetValue(Variable, k => context.Items[k], EvaluateAsNestedProperties);
 
--- a/NLog.Web.Tests/DefaultHttpContextAccessorTests.cs
+++ b/NLog.Web.Tests/DefaultHttpContextAccessorTests.cs
@@ -1,0 +1,29 @@
+﻿using System.IO;
+using System.Text;
+using System.Web;
+using Xunit;
+
+namespace NLog.Web.Tests
+{
+    public class DefaultHttpContextAccessorTests
+    {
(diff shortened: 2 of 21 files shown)


304NotModified:
Looks really nice! No comments :)

epignosisx:
I'm going to need to take a dependency on a mocking library in order to mock HttpContext. I looked at NLog/NLog but we are not using one. Do you have any preference?

304NotModified:
Sorry about the delay.

I prefer Moq, but I think NSubstitute is also fine.

epignosisx:
I'll do Moq then. Moq used to be my favorite but the simplicity of NSubstitute won me over.

304NotModified:
Well if NSubstitute is in your opinion is a lot easier to read/write, maybe that's a better choice. (because I prefer that also others can add unit tests easily) :)

epignosisx:
@304NotModified done with all the tasks. Let me know what you think.

NLog.Web/LayoutRenderers/AspNetItemValueLayoutRenderer.cs
```diff
@@ -56,18 +56,14 @@ public class AspNetItemValueLayoutRenderer : LayoutRenderer
         /// </summary>
         /// <param name="builder">The <see cref="StringBuilder"/> to append the rendered data to.</param>
         /// <param name="logEvent">Logging event.</param>
-        protected override void Append(StringBuilder builder, LogEventInfo logEvent)
+        protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
```
304NotModified: I'm doubting if this is backwards-compatible when we create a subclass of AspNetItemValueLayoutRenderer. There are two cases:

1 ) override the full behavior: before override `Append`.  Now we need to override DoAppend
2) Call `base.Append` - this still works, this has now also other behaviour

What do you think?

304NotModified:
Impressive! Just one note about backwards-compatibility.


Really good work!

image

https://codecov.io/github/NLog/NLog.Web/NLog.Web/LayoutRenderers?ref=aec90a22a34670165d5f1ca134591f6f46c6cd43

NLog.Web/LayoutRenderers/AspNetItemValueLayoutRenderer.cs
epignosisx: Maybe I'm missing something, but I dont see a backward-compatibility issue. Any subclass that calls base.Append will still experience the same behavior, the implementation changed but the behavior remains the same. If the subclass doesn't call base.Append then it is opting out of the base class behavior.
The DoAppend method is optional to override, but I would add to the docs that is the recommended way to write aspnet layout renderers that rely on the HttpContext since it makes the layout renderers more unit test friendly.

epignosisx:
@304NotModified have u had time to review my comment on your concern about backward compatibility.

304NotModified:
@epignosisx will look at this now. Thanks for the reminder

304NotModified:
I made two examples, and indeed, both behave the same:

```c#
        private class AspNetSessionIDLayoutRenderer2 : AspNetSessionIDLayoutRenderer
        {
           
            protected override void Append(StringBuilder builder, LogEventInfo logEvent)
            {
                builder.Append("before");
                //before: AspNetSessionIDLayoutRenderer.append() 
                //after: AspNetLayoutRendererBase.append() -> AspNetSessionIDLayoutRenderer.DoAppend 
                base.Append(builder, logEvent);
                builder.Append("after");
            }
        }
```


```c#
        private class AspNetSessionIDLayoutRenderer3 : AspNetSessionIDLayoutRenderer
        {

            protected override void Append(StringBuilder builder, LogEventInfo logEvent)
            {
                builder.Append("fixed");
            }
        }
```

I needed a double check, because of semver.

Issue #21, opened 2016-01-08
Additional Request Layout renderers + split into new layout renderers


epignosisx:
What do you think about including AspNet layout renderers for the following:

1.  User Agent: useful to correlate an issue with an specific browser or better yet to identify that a bot performed the request and ignore the error since must likely the bot passed invalid data.
2. Request Url: super important to try to reproduce the error. What page generated the error!
3. Referer Url: useful when trying to understand how a user end up in the current url.
4. Post Body: critical to know what data the user entered that generated the error. Caveat: we have to provide a way to mask certain information like PII and PCI.
5. IP Address: useful for analytics. You can understand where your users are coming from.
6. True Client IP Address: if a website is behind a CDN like Akamai or CloudFlare the IP address the web app receives is from one of the nodes of the CDN, they usually send the true client IP in a HTTP header. Caveat: each CDN uses a different headers so we will need to make it configurable but have a few of the major players.

304NotModified:
Sounds nice! We can do this all now I think, but all in a generic approach (`${aspnet-request}`).

6 is a bit to detailed, but we can make a `${aspnet-request-header}`?

I also miss some MVC related stuff now (route name, controller, action). Do you think is appropriate to add them to this repos?

epignosisx:
You are right. Most of what I'm proposing can be obtained with {aspnet-request}. 
It is true (6) is too specific, but it can be done with ServerVariables, however, it is not very intuitive:

Say you send a header like this:

    X-Forwarded-For: 105.45.66.789

You would retrieve it like:

    Request.ServerVariables["HTTP_X_FORWARDED_FOR"]

Yes, prefixed with "HTTP_",  all uppercase, and dashes become underscores.

Creating a ${aspnet-request-header} would make it more intuitive, or we could also extend the ${aspnet-request} to have a **Header** property and look in **Request.Headers**.

======================

Personally I don't find the controller name and action name very useful. I can always infer it from the request url, but others may find them useful. The good thing is that the API to retrieve this information is not in **System.Web.Mvc**, but rather **System.Web.Routing**. I don't like the idea of taking a reference on **System.Web.Mvc** in NLog.Web. As a user, I wouldn't be happy to have a Web Forms only app, have NLog.Web put a System.Web.Mvc assembly in the bin folder.

epignosisx:
I don't think we can get Number 4 (Post Body) from the current layout renderers. We are able to access individual keys of the Request.Form but not the whole body as a string.

[...]

PR #27, opened 2016-02-03
Add ability to render Headers

petemounce:


304NotModified:
Sorry, made a merge conflict because of another waiting PR (https://github.com/NLog/NLog.Web/pull/22), which improved the coverage much!

304NotModified:
Thanks for this PR. I looks good. We can merge this after fix of the conflict, BUT...

In https://github.com/NLog/NLog.Web/issues/21 is discussed to move this to a separate layout renderer, e.g. `${aspnet-request-header}`. What do you think about that? It's maybe a bit inconsistent with the current code, but `aspnet-request:header=a:Item=b` also is a bit odd (unclear what the order is and no influence on it). What do you think?

petemounce:
@304NotModified I've rebased and adjusted the tests to use that introduced by #22.

I don't know about moving to a separate layout render for this - System.Web's days are numbered in the face of OWIN and ASP.NET vNext cutting loose. I think if a new layout renderer is introduced, it would be a better direction to introduce one that can reach into the [OWIN environment](http://owin.org/html/owin.html#32-environment) for information to render into the log. While we're on that - are you aware of [pysco68/Pysco68.Owin.Logging.NLogAdapter](https://github.com/pysco68/Pysco68.Owin.Logging.NLogAdapter)? Hi @pysco68!

304NotModified:
OK thanks! 

No I'm not aware of owin for this. No experience with that. Is that something we should implement?

304NotModified:
Maybe dump question, but are owin and aspnet5 related?

petemounce:
@304NotModified aspnet5 is built on top of OWIN. OWIN is basically an abstraction layer that allows a .NET website to be separated from the hosting - so it becomes much easier to host in, for example, one's own process (console app, windows service), in-memory (tests), or (still) IIS via a System.Web Adaptor. MS WebApi already supports OWIN, ASPNET MVC will do in the next release, and most of the non-MS OSS frameworks also already do.

OWIN gives a set of interfaces for requests, responses, and then also the notion that http is done via a chain-of-responsibility pipeline, which allows different middlewares to be plugged in that are agnostic of the web framework (Nancy, WebApi, soon MVC, etc) - this makes cross cutting concerns much easier.

petemounce:
thanks for the merge :)

petemounce:
IMO yes, OWIN is definitely something that it would be great to have NLog support.

304NotModified:
Thanks for the info @petemounce !

About OWIN, yes it's nice although I have no problems to keep it external. Creating 2 times the same thing, that hasn't my preference. If the author (@pysco68) likes to move it to an official NLog repos, that would be also possible. (but we have to discuss how about license, ownership, maintenance etc). 

Will publish 4.2 when I improved and merged https://github.com/NLog/NLog.Web/pull/28

pysco68:
Hello folks,

I didn't really understand the how my NLogAdapter got into this PR but anyhow I read the notification. I'm willing to discuss any option. As far as the code is considered, it's under MIT license so you can do pretty much whatever you want.

Just tell me when and where you'd like to discuss that, I believe that this PR isn't the right place for that however.

304NotModified:
@petemounce I released 4.2 with this change. #28 will be 4.3 (include coreCLR)

petemounce:
Thanks!

--

PR #30, opened 2016-02-07
Added ASP.NET 5 compatible package

### LINKED ISSUE #29: Make ASPNET 5 compatible
This library isn't working with ASP.NET 5. The `HttpContext.Current` is not there:

http://stackoverflow.com/questions/31243068/httpcontext-current-in-asp-net-vnext

Too bad, I think we need a new package for that? (e.g. NLog.Web.Aspnet5) - as I didn't found a way to detect ASP.NET5

304NotModified:
Fixes #29


304NotModified:
Will reopen after successful build

PR #37, opened 2016-02-17
fix syntax of readme.md

petemounce:
I think the JSON code block was improperly terminated - but there's a difference between the client-side and server-side markdown rendering, I think, so I cannot be certain without issuing the PR.


304NotModified:
Thanks!!

Issue #39, opened 2016-03-15
DNX / ASP 5 - ${aspnet-request:serverVariable result in System.NotSupportedException


sgutkin:
The documentation [https://github.com/NLog/NLog/wiki/AspNetRequest-layout-renderer](url)

implies that serverVariable  works for asp.net5 /coreclr but the source code

[https://github.com/NLog/NLog.Web/blob/master/NLog.Web.ASPNET5/LayoutRenderers/AspNetRequestValueLayoutRenderer.cs](url)

explicitly seems to throw an exception :

`else if (this.ServerVariable != null)`
`    { `
`    #if !DNX `
`        builder.Append(httpRequest.ServerVariables[this.ServerVariable]); `
`    #else `
`        throw new NotSupportedException(); `
`    #endi `

For example in my nlog.config, the first 3 parameters work fine but the last two throw internal nlog exceptions (Warn Exception in layout renderer. Exception: System.NotSupportedException: Specified method is not supported.)

- parameter name="@requestMethod" layout="${aspnet-item:variable=RequestMethod}"/
- parameter name="@responseParamThree" layout="${aspnet-request:header=X-Forwarded-For}"/
- parameter name="@responseParamFour" layout="${aspnet-request:serverVariable=REMOTE_ADDR}"/
- parameter name="@responseParamFive" layout="${aspnet-request:serverVariable=ALL_RAW}"/

From project.json

  "dependencies": {
.....
    "NLog": "4.4.0-beta1",
    "NLog.Config": "4.3.0-beta1",
    "NLog.Extensions.Logging": "1.0.0-rc1-final-2016-03-11",
    "NLog.Schema": "4.3.0-beta1",
    "NLog.Web.ASPNET5": "4.2.1",
.......

304NotModified:
That's sloppy documentation.  We can't read the server variables in dnx (or AFAIK)

sgutkin:
Yeah I figured as much.   In RC1 the remote IP isn't readily accessible on  Microsoft.AspNet.Http.HttpContext.  I think they plan to fix it for RC2.

304NotModified:
superseded by https://github.com/NLog/NLog.Web/issues/44

PR #40, opened 2016-04-02
DNX: Request renderer: Remove ServerVariable from API in ASP.NET5

304NotModified:
Remove the option from API as ServerVariables are non-existing in ASP.NET 5.

This will give an configexception instead of runtime. 

Fixes https://github.com/NLog/NLog.Web/issues/39


304NotModified:
breaking change?

304NotModified:
fixed in https://github.com/NLog/NLog.Web/pull/59

Issue #31, opened 2016-02-08
${aspnet-request}: could write some warnings when request isn't available 


komlachsv:
If don't check this:
```
context.Session != null
```

in method `AspNetRequestValueLayoutRenderer::Append`

if try write to log from Global.asax.cs on start application, we have warning: 

``` 
Exception in layout renderer: System.Web.HttpException (0x80004005): 
   Request is not available in this context

```

304NotModified:
Hey,

can you give some more info? You get an exception? Can you post relevant config and c# code?

I don't understand this issue as [AspNetRequestValueLayoutRenderer](https://github.com/NLog/NLog.Web/blob/master/NLog.Web/LayoutRenderers/AspNetRequestValueLayoutRenderer.cs) isn't using `Session`

304NotModified:
@komlachsv can you give us some more info?


related: http://stackoverflow.com/questions/2518057/request-is-not-available-in-this-context

304NotModified:
We can't prevent this. If `throwExceptions` is set to `false`, this exception is only logged to the internallogger.

vegar:
What other implications will setting `throwExceptions` to `false` have?
I'm afraid it will hide some potential miss configurations etc.

Would it be an option to wrap the fetching of Request in a utility method, trapping any HttpException trown?

304NotModified:
> What other implications will setting throwExceptions to false have?
> I'm afraid it will hide some potential miss configurations etc.


I hide all errors, so logging could fail without knowing it. But that's a good case? I don't like crashing applications because logging failed. 
Note: 
You can set 'throwExceptions=false` and `throwConfigExceptions=true`. see http://nlog-project.org/2016/04/16/nlog-4-3-has-been-released.html

Also you can check the internalLogger.

Xender:
How about just catching the System.Web.HttpException from the property access?
Especially that there is already a [null check](https://github.com/NLog/NLog.Web/blob/63966e5ef6a168577a1d4aab70b65921abaceffa/NLog.Web.ASPNET5/LayoutRenderers/AspNetRequestValueLayoutRenderer.cs#L79-L83) in place.

I suppose one would have to give up type inferrence and declare `httpRequest` variable first, then surround the assignement which does property access in a try...catch block, then proceed normally?
So the only non-trivial part seems to be giving up the type inference, and even this looks straight-forward - is this a problem?

vegar:
[...]

PR #56, opened 2016-05-11
Suppressing exception thrown when accessing HttpContext.Request

vegar:
As a response to https://github.com/NLog/NLog.Web/issues/31

I have *not* kept the old behavior for .net core.

### DIFF
--- a/NLog.Web.ASPNET5/LayoutRenderers/AspNetRequestValueLayoutRenderer.cs
+++ b/NLog.Web.ASPNET5/LayoutRenderers/AspNetRequestValueLayoutRenderer.cs
@@ -76,7 +76,7 @@ namespace NLog.Web.LayoutRenderers
         /// <param name="logEvent">Logging event.</param>
         protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
         {
-            var httpRequest = HttpContextAccessor.HttpContext.Request;
+            var httpRequest = HttpContextAccessor.HttpContext.TryGetRequest();
             if (httpRequest == null)
             {
                 return;
@@ -134,6 +134,21 @@ namespace NLog.Web.LayoutRenderers
 #else
                 builder.Append(httpRequest.HttpContext.Items[this.Item]);
 #endif
+            }
+        }
+    }
+
+    internal static class RequestAccessor
+    {
+        internal static HttpRequestBase TryGetRequest(this HttpContextBase context)
+        {
+            try
+            {
+                return context.Request;
+            }
+            catch (HttpException)
+            {
+                return null;
             }
         }
     }
--- a/NLog.Web.ASPNET5/LayoutRenderers/AspNetRequestValueLayoutRenderer.cs
+++ b/NLog.Web.ASPNET5/LayoutRenderers/AspNetRequestValueLayoutRenderer.cs
@@ -141,6 +141,7 @@ namespace NLog.Web.LayoutRenderers
 
     internal static class RequestAccessor
     {
+#if !DNX
         internal static HttpRequestBase TryGetRequest(this HttpContextBase context)
         {
             try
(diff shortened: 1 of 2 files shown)


NLog.Web.ASPNET5/LayoutRenderers/AspNetRequestValueLayoutRenderer.cs
```diff
@@ -137,4 +137,19 @@ protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
...
+    {
+        internal static HttpRequestBase TryGetRequest(this HttpContextBase context)
+        {
+            try
+            {
+                return context.Request;
+            }
+            catch (HttpException)
+            {
+                return null;
```
304NotModified: At least a InternalLogger.debug(ex)  can be added?

304NotModified:
Thanks!

NLog.Web.ASPNET5/LayoutRenderers/AspNetRequestValueLayoutRenderer.cs
vegar: The compiler refuses to accept `InternalLogger.Debug(ex)`. It will only accept two variants of the Debug method: One taking a string, and the other taking a string and a argument array. Is that as excepted?

304NotModified: Ah yes that's is added to NLog 4.3. 

Debug(Ex.tostring()) is preferred.

vegar: (y) With or without any further explanation?

`InternalLogger.Debug("Exception thrown when accessing Request: " + ex);`
or 
` InternalLogger.Debug(ex.ToString());

Do you want a separate test verifying that it is logged? 

Current test passes based on the default log level for the internal logger discarding debug statements. A separate test can be added to verify that the internal logger do pick up on the exception if log level is set appropriately.

304NotModified: The first please :) 

No need for testing.

vegar: image

304NotModified: LOL,

is this a response on "no need for testing"?

304NotModified: image

304NotModified:
coreclr build is a "bit" broken now ;)

NLog.Web.ASPNET5/LayoutRenderers/AspNetRequestValueLayoutRenderer.cs
vegar: Yes, I was taken by surprise.  It's not the response I'm used to ;-)

vegar:
I guess it's time to start looking into this .net core thing then...

vegar:
Is there a reason why the `IHttpContextAccessor` that is implemented inside of NLog.Web returns a `HttpContextBase` while the `IHttpContextAccessor` coming with .net core returns a `HttpContext` ?

304NotModified:
> s there a reason why the IHttpContextAccessor that is implemented inside of NLog.Web returns a HttpContextBase while the IHttpContextAccessor coming with .net core returns a HttpContext ?

AFAIK we need the `HttpRequestBase` in ASP.NET 4 (note: `HttpRequest` won't inherit from `HttpRequestBase ` !) and there is no `HttpRequestBase` in .NET CORE

vegar:
Ok. Glad you could use the code. Hopefully, we can delete our own implementations pretty soon :-)

Keep up the good work.

Page-Not-Found:
I have question about this PR. Wouldn't we want to implement the same TryGet Method for where we have access the Request?

PR #121, opened 2017-03-07
iis-site-name should return app name in ASP.NET Core

grokky1:
For ASP.NET Core:

`IHostingEnvironment.EnvironmentName` = "Development", "Production", "Staging", etc.
`IHostingEnvironment.ApplicationName` = the app name

The layout renderer for "iis-site-name" is incorrectly rendering the first instead of the second.

### DIFF
--- a/NLog.Web.AspNetCore/LayoutRenderers/IISInstanceNameLayoutRenderer.cs
+++ b/NLog.Web.AspNetCore/LayoutRenderers/IISInstanceNameLayoutRenderer.cs
@@ -36,7 +36,7 @@ namespace NLog.Web.LayoutRenderers
 
 #if NETSTANDARD_1plus
             var env = ServiceLocator.ServiceProvider?.GetService<IHostingEnvironment>();
-            builder.Append(env?.EnvironmentName);
+            builder.Append(env?.ApplicationName);
 
 #else
             builder.Append(HostingEnvironment.SiteName);


grokky1:
`IHostingEnvironment.EnvironmentName` is actually quite interesting for diagnostics purposes. It could be a useful separate layout renderer.

304NotModified:
Good catch!

Maybe we should add a new `LayoutRenderer` (e.g. "iis-site-env") if people where dependent on the EnvironmentName ?

grokky1:
Yeah that's a good idea. It's also useful for diagnostics.

I only have netstandard installed on my workstation though, so I don't know what would be the equivalent for MVC5. I'll try do it later.

304NotModified:
Great!

grokky1:
Looks like the [environment is a new thing in Core](http://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments). I don't believe it's available in pre-Core MVC.

All the layout renders seem to work with ASP.NET and ASP.NET Core because of:

    #if !NETSTANDARD_1plus
    foo
    #else
    bar
    #endif

How should I handle it?

304NotModified:
just do `#if NETSTANDARD_1plus` for the whole class (so it's only supported in ASP.NET Core)

grokky1:
Sorry I'm not really familiar with GitHub and have problems updating that repo.

1 So the fix to `IISInstanceNameLayoutRenderer.cs` is like above.

2 And below is the new file `AspNetEnvironmentLayoutRenderer.cs` for the environment name

    #if NETSTANDARD_1plus
    using System;
    using System.Text;
    using Microsoft.AspNetCore.Hosting;
    using Microsoft.Extensions.DependencyInjection;
    using NLog.LayoutRenderers;
    using NLog.Web.Internal;

    namespace NLog.Web.LayoutRenderers
    {

        /// <summary>
        /// Rendering development environment. <see cref="IHostingEnvironment"/>
        /// </summary>
        [LayoutRenderer("aspnet-environment")]
        // ReSharper disable once InconsistentNaming
        public class AspNetEnvironmentLayoutRenderer : LayoutRenderer
        {
            /// <summary>
            /// Append to target
            /// </summary>
            /// <param name="builder">The <see cref="StringBuilder"/> to append the rendered data to.</param>
            protected override void Append(StringBuilder builder, LogEventInfo logEvent)
            {
                var env = ServiceLocator.ServiceProvider?.GetService<IHostingEnvironment>();
                builder.Append(env?.EnvironmentName);
            }
        }
    }
    #endif

It is not possible (as far as I know) to reasily test this, because it needs a `IHostingEnvironment`, which you can get by running a fake webhost/server. Not sure you want that kind of heavy load in the tests.

304NotModified:
Do you need some help with git/github or no time for that?

grokky1:
@304NotModified I think I messed thing sup by trying to use GitHub's online editing features. Not sure how to go backwards 😆 

Summary:
- This PR works as intended to fix the bug.
- You said it's a good idea to also add a `AspNetEnvironmentLayoutRenderer`, which I had a problem with. But that is a separate matter. I did do it though and added the code in a comment above.

Might be a good idea to split this issue in two, because the bug should not be depedent on the new feature. As is stands, the layout renderer is producing the wrong data, which this PR fixes.

304NotModified:
thanks @grokky1 

moved it to https://github.com/NLog/NLog.Web/pull/125

grokky1:
@304NotModified Thanks for splitting it in two.

Just so I understand because I'm a little confused, that issue you just opened is for the new feature which logs the aspnet env name? And what about the bug fix in this issue for the aspnet app name... is it going to be included at some point (I recall the bugfix works and passed those tests)?

304NotModified:
> Just so I understand because I'm a little confused, that issue you just opened is for the new feature which logs the aspnet env name?

See commit (or diff), your commit is also included: https://github.com/NLog/NLog.Web/pull/125/commits

PR #133, opened 2017-05-01
Improve performance with many threads

304NotModified:
See https://github.com/NLog/NLog/issues/2064


snakefoot:
@304NotModified If the usage of ServiceLocator.ServiceProvider should be avoided. Then maybe these should also be fixed:

- AspNetEnvironmentLayoutRenderer.cs
- IISInstanceNameLayoutRenderer.cs

304NotModified:
thanks will check them!

snakefoot:
@304NotModified Wondering if it would make sense also to add:

```
protected override void CloseLayoutRenderer()
{
   _httpContextAccessor = null;
}
```

304NotModified:
good idea! Added to https://github.com/NLog/NLog.Web/issues/134

PR #150, opened 2017-06-09
Update vulnerable dependency

johnkors:
Hi,

I scanned Nlog.Web.AspNetCore using [dotnet retire](https://github.com/RetireNet/dotnet-retire) and found use of a vulnerable dependency. This PR bumps the package that brings this dependency in.

image

Dependency chain resulting in use [of vulnerable dependency `System.Text.Encodings.Web/4.0.0`](https://github.com/dotnet/announcements/issues/12) :

[Microsoft.AspNetCore.Routing.Abstractions/1.0.3](https://www.nuget.org/packages/Microsoft.AspNetCore.Routing.Abstractions/1.0.3)
=> 
[Microsoft.AspNetCore.Http.Abstractions/1.0.2](https://www.nuget.org/packages/Microsoft.AspNetCore.Http.Abstractions/1.0.2)
=>
[System.Text.Encodings.Web/4.0.0](https://www.nuget.org/packages/System.Text.Encodings.Web/4.0.0)

### DIFF
--- a/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
+++ b/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
@@ -36,12 +36,14 @@
     <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="1.0.2" />
     <PackageReference Include="Microsoft.AspNetCore.Http.Features" Version="1.0.2" />
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="1.0.2" />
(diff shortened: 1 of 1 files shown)


304NotModified:
ah cool thanks! 

Could you also analyse NLog.extensions.Logging and NLog? Or isn't that needed?

johnkors:
`NLog.Web.AspNetCore` references these other two projects, so AFAIK - any dependency they bring in would also be present for analysis (it checks the `/obj/project.assets.json` file of .NET Core projects). So I _think_ they're already checked :)

304NotModified:
oops,we broke the build but AppVeyor showing green.

PR #162, opened 2017-07-26
Update AssemblyVersionLayoutRenderer docs for .NET core

Valeriy1991:
Fixed comments for Name property (for .NET Full and Standard results of layout renderer will be different)


Valeriy1991:
@304NotModified done! Thanks.

Issue #172, opened 2017-08-31
Move config from Startup to Config


grokky1:
**TL;DR**
Below are some changes needed to make most projects work with ASP.NET Core2 (netcoreapp2.0). I've also submitted PRs for these changes.

---

There are a number of "please upgrade NLog to ASP.NET Core 2.0" issues, but this is a little different.

There are a few workarounds documented here to make NLog work on netcoreapp2.0. I have it working.

However, logging is now split over `Startup.cs` and `Program.cs`. In 1.1 all logging config was done in `Startup.Configure()`. But in 2.0 logging is configured as soon as the app starts, in `Program.BuildWebHost(string[])`. That makes logging available immediately, including from within the startup code.

Right now default logging is configured in `Program`, and NLog is configured later in `Startup.Configure()`. It should be moved into `Program`.

Here is an example of something that doesn't work because of this. I do this to log app lifetime events:

````csharp
public Startup(IConfiguration configuration, ILogger<Startup> logger)
{
  _configuration = configuration;
  _logger = logger;
}

private readonly IConfiguration _configuration;
private readonly ILogger<Startup> _logger;

public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory, IApplicationLifetime appLifetime)
{
  // other stuff...
  appLifetime.ApplicationStarted.Register(Started);
  appLifetime.ApplicationStopping.Register(Stopping);
  appLifetime.ApplicationStopped.Register(Stopped);
  // other stuff...
  // configure NLog in the usual way
}

private void Started()
{
  _logger.LogInformation("App started");
}

private void Stopping()
{
  _logger.LogInformation("App stopping");
}

private void Stopped()
{
  _logger.LogInformation("App stopped");
}
````

The `Started` event is called, and it logs to the Console and Debugger loggers. But not to NLog. That's because NLog was configured too late.

To make it work - instead of injecting `ILogger<Startup>` into the constructor, I can do this in `Configure()` after setting up NLog: `_logger = loggerFactory.CreateLogger<Startup>();`.

As you can see this is messy. All config must be in `Program`.

This can be hotfixed easily. In the `Program.BuildWebHost(string[])` we could add this:

````csharp
return WebHost
  .CreateDefaultBuilder(args)
  .ConfigureLogging((context, builder) => {
    context.HostingEnvironment.ConfigureNLog("nlog.config");
    builder.AddNLog();         // requires FIX 1
[...]

PR #173, opened 2017-08-31
Updated to support netcoreapp2.0

grokky1:
Related to [NLog.Web 172](https://github.com/NLog/NLog.Web/issues/172)

### DIFF
--- a/NLog.Web.AspNetCore/AspNetExtensions.cs
+++ b/NLog.Web.AspNetCore/AspNetExtensions.cs
@@ -23,9 +23,9 @@ namespace NLog.Web
         /// </summary>
         /// <param name="app"></param>
 
-        public static void AddNLogWeb(this IApplicationBuilder app)
+        public static void AddNLogWeb(this ILoggingBuilder builder)
         {
-            ServiceLocator.ServiceProvider = app.ApplicationServices;
+            ServiceLocator.ServiceProvider = builder.Services;
         }
 
         /// <summary>


grokky1:
Discussion point: should we not take this opportunity to simplify things?

If we do this

````csharp
public static void AddNLogWeb(this ILoggingBuilder builder)
{
  builder.AddNLog();                                                 // NEW
  ServiceLocator.ServiceProvider = builder.Services;
}
````

Then instead of 
````csharp
context.HostingEnvironment.ConfigureNLog("nlog.config");
builder.AddNLog();
builder.AddNLogWeb();
````

We can rather have this simpler config
````csharp
context.HostingEnvironment.ConfigureNLog("nlog.config");
builder.AddNLogWeb();
````

xperiandri:
@grokky1, yep, looks better.

304NotModified:
> @grokky1, yep, looks better.

Indeed. This was a historical things (breaking change).

This is also a breaking change isn't? Can't we add the new method next to it and mark the other one obsolete?

grokky1:
@304NotModified Yes you're right, we should mark the old one obsolete. However! If people still make both calls, then it would call the config code twice. So we need to 1) think of a way around that, or 2) introduce a new name e.g. `AddNLogCore()` or somthing like that... yuck.

However, we can't go ahead with this until we deal with the extensions PR first: https://github.com/NLog/NLog.Extensions.Logging/pull/126

304NotModified:
> @304NotModified Yes you're right, we should mark the old one obsolete. However! If people still make both calls, then it would call the config code twice. So we need to 1) think of a way around that, or 2) introduce a new name e.g. AddNLogCore() or somthing like that... yuck.

Nothing goes wrong when calling them both, so not a problem :)

304NotModified:
> However, to proceed we must first deal with the NLog.Extensions PR: NLog/NLog.Extensions.Logging#126

Yes, wasn't there yet. :) Thanks for the reminder.

304NotModified:
superseded by #181

Issue #157, opened 2017-06-21
Getting-started-with-ASP.NET-Core vs2017 


NelsonRothermel:
I followed https://github.com/NLog/NLog.Web/wiki/Getting-started-with-ASP.NET-Core-(csproj---vs2017) for my VS2017 .NET Core project.  Step 4 shows the following:

```
public Startup(IHostingEnvironment env)
{
    env.ConfigureNLog("nlog.config");
}

public void ConfigureServices(IServiceCollection Services)
{
    //call this in case you need aspnet-user-authtype/aspnet-user-identity
    services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
}
```

That doesn't work for me since `Startup` runs before `ConfigureServices`.  When `ConfigureNLog` runs I get the following error: `2017-06-20 21:39:37.7335 Warn Missing IHttpContextAccessor. Has it been registered before loading NLog Configuration? Consider reloading NLog Configuration after having registered the IHttpContextAccessor.`

I fixed it by moving `ConfigureNLog` inside of the `Configure` method and taking a `IHostingEnvironment` parameter, but was wondering if the documentation was incorrect or if the code was intended to be used that way somehow.

mhamri:
i moved the `ConfigureNLog` as you said, but still i'm getting the said error

304NotModified:
Is this on ASP.NET Core 2?

mhamri:
I'm using asp.Net core 1.2.

But this message is actually wrong.  if i log inside a class that doesn't
have action and controller,  then this message will appear.

304NotModified:
> if i log inside a class that doesn't
> have action and controller,  then this message will appear.

could you please give an example? I'm not fully get it, thanks!

mhamri:
```c#
public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory, ApplicationDbContext context, IServiceProvider services, Paths pathHelper)
{
    ILogger<Startup> logger = null;
    if (loggerFactory != null)
    {
        //loggerFactory.AddConsole(Configuration.GetSection("Logging"));
        //loggerFactory.AddDebug();

        loggerFactory.AddNLog();


        app.AddNLogWeb();
        env.ConfigureNLog("nlog.config").Reload(); 
        /*
        i also have env.ConfigureNLog("nlog.config") at the startup constructor, 
       but still after `addNlogWeb`  i'm trying to configure nlog and reload it,
[...]

PR #180, opened 2017-09-29
Fix IHttpContextAccessor warning

304NotModified:
fixes https://github.com/NLog/NLog.Web/issues/157

### DIFF
--- a/NLog.Web.AspNetCore/LayoutRenderers/AspNetLayoutRendererBase.cs
+++ b/NLog.Web.AspNetCore/LayoutRenderers/AspNetLayoutRendererBase.cs
@@ -40,19 +40,18 @@ namespace NLog.Web.LayoutRenderers
         /// <returns>HttpContextAccessor or <c>null</c></returns>
         public IHttpContextAccessor HttpContextAccessor
         {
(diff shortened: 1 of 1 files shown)


snakefoot:
@304NotModified Could be nice if AspNetLayoutRendererBase.Initialize() generated a warning, when HttpContextAccessor is null (And not a debug message for every property-access). As the initial proposal suggested #133 (but changed during refactoring).

Maybe also move the logic in the AspNetLayoutRendererBase-constructor into AspNetLayoutRendererBase.Initialize().

304NotModified:
> @304NotModified Could be nice if AspNetLayoutRendererBase.Initialize() generated a warning, when HttpContextAccessor is null (

That was the case and we got multiple tickets for the warning. So now debug and only on usage.

snakefoot:
Did you notice comments like these:

> All logging fields like aspnet-* are not included in logs.

Which actually means the logging is not working, because it has been initialized wrong. Maybe the NLog Example should be fixed to perform proper initialization so the layout-renderer works.

304NotModified:
The new way (working on it), is less difficult to setup :)

Surge001:
Can't wait on the new way to be released. Please let us know in this trail when released! 
thnx!

304NotModified:
Yes I will! Hopefully this weekend!

Issue #187, opened 2017-09-30
Make `  <add assembly="NLog.Web.AspNetCore"/>` unneeded


304NotModified:
Make 

```
  <extensions>
    <add assembly="NLog.Web.AspNetCore"/>
  </extensions>

```

Unneeded in the config

304NotModified:
we need `RegisterItemsFromAssembly`  on `ConfigurationItemFactory`

304NotModified:
doesn't work if we do logmanager.configuration.reload()

304NotModified:
fixed for asp.net core 2

Issue #184, opened 2017-09-30
Simplify setting up the NLog config 


304NotModified:
Simplify setting up NLog for ASP.NET Core 2

Currently setting up takes a lot of steps. It should be as easy as possible  - but also flexible enough. Also the current design not really suitable for ASP.NET Core 2

Current config (code set-up only):

```c#
using NLog.Extensions.Logging;
using NLog.Web;

public Startup(IHostingEnvironment env)
{
    env.ConfigureNLog("nlog.config");
}

public void ConfigureServices(IServiceCollection Services)
{
    //call this in case you need aspnet-user-authtype/aspnet-user-identity
    services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
}

// This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
{

    //add NLog to ASP.NET Core
    loggerFactory.AddNLog();

    //add NLog.Web
    app.AddNLogWeb();

   //note: remove the old loggerFactory, like loggerFactory.AddConsole and  loggerFactory.AddDebug
```

Redesign the API design for ASP.NET Core 2

(maybe also replace `ConfigureNLog(this IHostingEnvironment env`)

304NotModified:
@grokky1 

I think it will be something like this, do you think this is simple enough?

```c#
  public class Program
    {
        public static void Main(string[] args)
        {
            // NLog: setup the logger first to catch all errors
            LogManager.Configuration = new XmlLoggingConfiguration("NLog.config");
            var logger = LogManager.GetCurrentClassLogger();
            try
            {
                BuildWebHost(args).Run();
            }
            catch (Exception e)
            {
                //NLog: catch setup errors
                logger.Error(e, "Stopped program because of exception");
                throw;
            }
        }

        public static IWebHost BuildWebHost(string[] args) =>
            WebHost.CreateDefaultBuilder(args)
                .UseStartup<Startup>()
                .UseNLog() // NLog: setup NLog for Dependency injection
                .Build();
    }

```

304NotModified:
i'm doubting to create a helper for 

```C#
LogManager.Configuration = new XmlLoggingConfiguration("NLog.config");
var logger = LogManager.GetCurrentClassLogger();
```

But maybe it's needed (for #187 at least)

grokky1:
[...]

PR #191, opened 2017-10-07
Added  NLogBuilder and  ASP.NET Core 2 example

304NotModified:
- added `NLogBuilder.ConfigureNLog`, https://github.com/NLog/NLog.Web/issues/184
- added `NLogAspNetCoreOptions`
- auto register:  `<extensions>` not needed anymore in the config, fixes https://github.com/NLog/NLog.Web/issues/187
- fix namespaces, 
- updated docs
- added ASP.NET Core 2 example (build on AppVeyor)


examples/ASP.NET Core 2/Visual Studio 2017/ASP.NET Core 2 - VS2017/Program.cs
```diff
@@ -0,0 +1,41 @@
...
+using NLog.Web;
+
+namespace ASP.NET_Core_2___VS2017
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            // NLog: setup the logger first to catch all errors
+            var logger = NLogStart.InitLogger("NLog.config");
```
304NotModified: @snakefoot inspiration for a better names?

snakefoot: InitLogger is very confusing. It almost seem like you are creating a logger named "NLog.config".

Maybe rename NLogStart to NLogBuilder, and rename "InitLogger" to "LoadConfig" where LoadConfig returns NLogBuilder (fluent interface like)

NLogBuilder.LoadConfig("NLog.config").GetCurrentClassLogger()

NLogBuilder.GetCurrentClassLogger()

snakefoot: Be aware that calling GetCurrentClassLogger within NLogBuilder will capture the wrong classname.

304NotModified: > NLogBuilder.LoadConfig("NLog.config").GetCurrentClassLogger()

`LoadConfig` could return the `LogFactory`, not sure it that's clear. What do you think?

snakefoot: > not sure it that's clear. What do you think?

Actually what you want is just the LogManager-interface, but you want people to perform the registration of AspNetExtensions before loading the config (From file or programmatically). Think it would be easier for all NLog-users if LogManager remained the default goto-place.

Think you want something like ConfigureNLog that just configures everything, and then people can use LogManager from there.

Maybe you should just change the RegisterExtendedItems() to attempt loading known good NLog-assemblies. Like NLog.Extended.dll and NLog.Web.dll and NLog.Web.AspNetCore.dll

Usually this auto-loading of NLog-dlls works out-of-the-box, but not when using Nuget-package-dlls that reside in different sub-folder.

snakefoot: Maybe better that you just call it ConfigureNLog, but maybe you should change all ConfigureNLog-implementations to return the LogFactory object (instead of LoggingConfiguration). Since the LoggingConfiguration-object is kind of dull, and you can still access it through LogFactory.

304NotModified: > Since the LoggingConfiguration-object is kind of dull, and you can still access it through LogFactory.

What's dull about it? That changing the config won't work until you reset the .config property?

304NotModified: well Flush isn't needed in this case? Or do you think it's wise to add it after `BuildWebHost`?

snakefoot: I can just see that people like the fluent-interface of Serilog, and it requires that you have access to a "powerful" object like the LogFactoy. But yes normally one will just be calling GetCurrentClassLogger().

But I prefer to have 3 stages. Startup - Running - Shutdown. Where Startup is where I load the NLog-config and the Running-stage where I acquire my loggers. Each stage happens at different locations in the code, so I don't need to build the nlog-config and get the logger-interface at the same location.

304NotModified: We load it once here?

Fluent config would be great indeed. With the json config feature, that's one of the next steps for 4.6/4.7 I think

snakefoot: > We load it once here?

Yes but one might think that if I need to get a logger at different place, then I should just do the same (Call LoadConfig/ConfigureNLog and then GetCurrentClassLogger), and then one might be surprised by the side-effects of the config-reload.

Or one might be cheated by the NLog-singleton-behavior, as it is hard to have multiple configs at the same time.

examples/ASP.NET Core 2/Visual Studio 2017/ASP.NET Core 2 - VS2017/Program.cs
```diff
@@ -0,0 +1,43 @@
...
+using NLog.Web;
+
+namespace ASP.NET_Core_2___VS2017
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            // NLog: setup the logger first to catch all errors
+            var logger = NLogBuilder.ConfigureNLog("NLog.config").GetCurrentClassLogger();
```
304NotModified: @snakefoot updated it! I think is indeed a lot better.

@grokky1 any opinion on this? Do you think this is clear?

examples/ASP.NET Core 2/Visual Studio 2017/ASP.NET Core 2 - VS2017/Program.cs
```diff
@@ -0,0 +1,41 @@
...
```
304NotModified: @snakefoot updated it! I think is indeed a lot better.

@grokky1 any opinion on this? Do you think this is clear?

grokky1: The latest changes are good. Looking forward to using this.

grokky1: @304NotModified Sorry busy wrapping up a sprint so didn't get a chance, will try soon!!

PR #273, opened 2018-04-09
Added public method to configure the NLog ServiceLocator

### LINKED ISSUE #270: custom provider for NLog doesn't wire up aspnet web layouts due to ServiceLocator pattern
In looking at the core code of the extension method UseNLog, I discovered there is an internal class called ServiceLocator where it sets the ServiceProvider, based on IServiceCollection during initialization.

I need to make a custom version of the service provider as I want to do some things that I don't believe are supported (creating a different loglevel on the logger based on a header value).  If I add the code for UseNLog and my code to add my provider, those missing log attributes get added, but there are two loggers now.. And I get duplicate logs.

If I don't call UseNLog, I get this error from the internal logger:(https://github.com/NLog/NLog.Web/blob/master/NLog.Web.AspNetCore/LayoutRenderers/AspNetLayoutRendererBase.cs)

Missing serviceProvider, so no HttpContext.

The attributes do not work correctly as they have a dependency on the ServiceLocator ServiceProvider being populated..

        <attribute name="requestIp" layout="${aspnet-Request-IP}"/>
        <attribute name="controller" layout="${aspnet-MVC-Controller}"/>
        <attribute name="action" layout="${aspnet-MVC-Action}"/>
        <attribute name="identity" layout="${identity}"/>

Please assist..

snakefoot:
Attempt to resolve  #270

### DIFF
--- a/NLog.Web.AspNetCore/AspNetExtensions.cs
+++ b/NLog.Web.AspNetCore/AspNetExtensions.cs
@@ -30,9 +30,7 @@ namespace NLog.Web
 #endif
         public static void AddNLogWeb(this IApplicationBuilder app)
         {
-            ServiceLocator.ServiceProvider = app.ApplicationServices;
-            ConfigurationItemFactory.Default.RegisterItemsFromAssembly(typeof(AspNetExtensions).GetTypeInfo().Assembly);
-            LogManager.AddHiddenAssembly(typeof(AspNetExtensions).GetTypeInfo().Assembly);
+            app.ApplicationServices.SetupNLogServiceLocator();
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


NLog.Web.AspNetCore/AspNetExtensions.cs
```diff
@@ -125,5 +124,20 @@ public static IWebHostBuilder UseNLog(this IWebHostBuilder builder, NLogAspNetCo
...
+        /// </summary>
+        /// <remarks>
+        /// Should only be used if the standard approach for configuring NLog is not enough
+        /// </remarks>
+        /// <param name="serviceProvider"></param>
+        public static IServiceProvider SetupNLogServiceLocator(this IServiceProvider serviceProvider)
+        {
+            ServiceLocator.ServiceProvider = serviceProvider;
+            ConfigurationItemFactory.Default.RegisterItemsFromAssembly(typeof(AspNetExtensions).GetTypeInfo().Assembly);
+            LogManager.AddHiddenAssembly(typeof(AspNetExtensions).GetTypeInfo().Assembly);
```
304NotModified: Maybe reuse gettypeinfo/assembly for both calls?

304NotModified:
Thanks!

PR #283, opened 2018-04-28
Avoid assembly lookup on static class initialization

### LINKED ISSUE #281: NLog.Web.Aspnetcore 4.5.2 error in ServiceFabric
The NLog.Web.Aspnetcore 4.5.2 version in ServiceFabric will cause the service to be unable to start, the local debug mode is running, the formal release online version will appear, and I doubt it is a file path problem.

snakefoot:
Guessing that Assembly lookup during startup is not welcome. Trying to resolve #281 

Another example of assembly lookup during startup is failing for ServiceFabric: https://github.com/quartznet/quartznet/issues/619 (Though here the bug is in `SerilogLogProvider.IsLoggerAvailable`)

### DIFF
--- a/NLog.Web.AspNetCore/AspNetExtensions.cs
+++ b/NLog.Web.AspNetCore/AspNetExtensions.cs
@@ -21,8 +21,6 @@ namespace NLog.Web
     /// </summary>
     public static class AspNetExtensions
     {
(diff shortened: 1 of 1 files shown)


304NotModified:
Thanks!

PR #132, opened 2017-04-09
Fix invalid JSON format for ${aspnet-request-cookie} and ${aspnet-request-querystring}, added extra formatting options, consistent rendering of multiple values

304NotModified:
No breaking changes in config, but change in output of multiple values. It's now consistent and configurable. 

fixes https://github.com/NLog/NLog.Web/issues/119

todo:

- [x] option to add newline? -> separator options
- [x] option to join multiple cookie values (non core) -> no multivalue cookies
- [x] option for quotes? -> maybe later
- [x] separator flat, `:` or `=`  ->  option

PR #295, opened 2018-06-22
Value only aspnet request cookie

### LINKED ISSUE #294: aspnet-request-cookie Value Only
**Type**: Bug (documentation) / Feature Request

**NLog version**: 4.5.6

**Platform**: ASP.NET Core 2.0

**Current NLog config**:

```
layout="${aspnet-request-cookie:CookieNames=TestCookie}"
```

In case of a BUG:

- What is the current result?
Will print
```
TestCookie=TheCookieValue
```

- What is the expected result?
As per [documentation](https://github.com/NLog/NLog/wiki/AspNetRequest-Cookie-Layout-Renderer), value of the cookie *not including its name* written to the log.
```
TheCookieValue
```

- Did you checked the [Internal log](https://github.com/NLog/NLog/wiki/Internal-Logging)? No, not an error.

- Please post full exception details (message, stacktrace, inner exceptions): N/A

- Are there any workarounds? No

- Is there a version in which it did work? Don't Know

- Can you help us by writing an unit test? Absolutely

In case of a FEATURE REQUEST:
 
If this is a feature that should be added, perhaps as a flag like `ValueOnly`...

- Why do we need it?
Our logs are parsed and having the key/value pair syntax causes a problem for this. We need the value only if possible.

I'm happy to add this feature if that is OK.

alexangas:
Fixes #294 by adding a "ValuesOnly" option to the ${aspnet-request-cookie} renderer, with associated unit tests.

Also:
* clarifies naming of some methods/parameters where the word "values" is used
* fixes some typos / inconsistent spacing

### DIFF
--- a/NLog.Web.AspNetCore.Tests/LayoutRenderers/AspNetCookieLayoutRendererTests.cs
+++ b/NLog.Web.AspNetCore.Tests/LayoutRenderers/AspNetCookieLayoutRendererTests.cs
@@ -32,7 +32,6 @@ namespace NLog.Web.Tests.LayoutRenderers
     {
         public AspNetCookieLayoutRendererTests() : base()
         {
-
         }
 
         [Fact]
@@ -60,8 +59,6 @@ namespace NLog.Web.Tests.LayoutRenderers
             Assert.Empty(result);
         }
 
-
-
         [Fact]
         public void KeyNotFoundRendersEmptyString_Json_Formatting()
         {
@@ -75,7 +72,7 @@ namespace NLog.Web.Tests.LayoutRenderers
         }
 
         [Fact]
-        public void KeyFoundRendersValue_Cookie_Mulitple_Items_Flat_Formatting()
+        public void KeyFoundRendersValue_Cookie_Multiple_Items_Flat_Formatting()
         {
 #if ASP_NET_CORE
             //no multivalue keys in ASP.NET core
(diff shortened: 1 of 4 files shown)


alexangas:
The unit tests are failing because the implementation of GetCookies is different in the renderer between Core and non-Core. I tried to find out what and why but not sure what's happening here.

Is it still necessary to have two different implementations and therefore behaviours?

304NotModified:
> Is it still necessary to have two different implementations and therefore behaviours?

yes, we prefer to have it implemented on both platforms. Both are maintained.

alexangas:
I've investigated further but still not sure why they are different, I think the clue [is here](https://github.com/NLog/NLog.Web/pull/132/files#diff-c063a101745bdf6d32847793bb356b5fR81). However it seems to be the Core version that is missing something compared to Full, so in a separate PR I will attempt to fix this functionality in Core and also improve the unit testing side. Then I will finish this PR off. :smile:

alexangas:
As mentioned in previous comment, will come back to this after #299.

304NotModified:
need any help on this? :)

alexangas:
I've completed most of it, but need to verify and correct the behaviour of multi-values. I will use the full framework example app recently merged in to do this. Expect to have it completed over the next week sometime.

PR #296, opened 2018-06-22
Assemblyversion share web

### LINKED ISSUE #293: Features from NLog AssemblyVersionLayoutRenderer
**Type**: Bug (unexpected behaviour)

**NLog version**: 4.5.6

**Platform**: .NET Core 2 

**Current NLog config**: N/A

I implemented nlog/nlog#2480 expecting to be able to use the features added in an ASP.NET Core 2.0 application. After the official release came out, I noticed it didn't appear to be working correctly but only recently had time to investigate. After many hours I found that it is being overridden by the AssemblyVersionLayoutRenderer in this project. So documentation bug - this needs to be clear in the README! 😅 

However the more important question! Is it possible to address this override instead to either:

1. Inherit from the NLog AssemblyVersionLayoutRenderer, so that these features can be picked up. As part of the aforementioned work I split out the methods so it shouldn't take much work to be able to inherit from it (but would require minor code changes).

2. Move the logic for retrieving the assembly from HTTP context into NLog and remove from this project altogether so it is only in one place?

I'm happy to work on this.

alexangas:
Fixes #293 by copying logic from NLog

### DIFF
--- a/NLog.Web.AspNetCore.Tests/LayoutRenderers/AssemblyVersionLayoutRendererTests.cs
+++ b/NLog.Web.AspNetCore.Tests/LayoutRenderers/AssemblyVersionLayoutRendererTests.cs
@@ -13,9 +13,25 @@ namespace NLog.Web.AspNetCore.Tests.LayoutRenderers
         [Fact]
         public void AssemblyNameVersionTest()
         {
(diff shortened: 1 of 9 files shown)


alexangas:
This will be redone using Option 1 described in #293. I will aim to get this done tonight.

304NotModified:
👍

PR #297, opened 2018-06-26
Inheritance from AssemblyVersionLayoutRenderer

### LINKED ISSUE #293: Features from NLog AssemblyVersionLayoutRenderer

alexangas:
Fixes #293 and NLog/NLog#2783.

Initially this cannot be merged in as builds will fail until a NuGet package (pre-release or otherwise) is released that contains the code from NLog/NLog#2788 and the dependency herein is updated.

Tested against ASP.NET Core 2.x. Updated existing config with a specific NLog target to thoroughly test assembly version.

Could not test with ASP.NET Full framework due to issues with Visual Studio and non-Core projects. Should be able to do this by adding an example project targeting Full once this PR can build.


304NotModified:
FYI: I'm on holiday now, but I will try to release NLog 4.5.7 soon, so this PR will succeed.

304NotModified:
NLog 4.5.7 is now live :)

alexangas:
I've screwed up rebasing/merging here, so creating a new branch.

PR #301, opened 2018-07-09
Added ASP.NET Full Framework example project (.NET 4.6.1 - VS2017)

alexangas:
Used to test #293 and can also be used for testing other in-process scenarios that unit tests cannot easily or cannot at all cover.

Updated paths in Core test project nlog.config so they do not conflict and are clearly differentiated.


304NotModified:
Thanks!

PR #314, opened 2018-09-16
Better missing httpcontext reporting & Skip null-check of HttpContextAccessor.HttpContext in DoAppend 

snakefoot:
Has been checked

### DIFF
--- a/NLog.Web.AspNetCore/LayoutRenderers/AspNetRequestQueryStringLayoutRenderer.cs
+++ b/NLog.Web.AspNetCore/LayoutRenderers/AspNetRequestQueryStringLayoutRenderer.cs
@@ -1,16 +1,15 @@
-﻿using System.Text;
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
 #if !ASP_NET_CORE
-using System.Web;
 using System.Collections.Specialized;
+using System.Web;
 #else
 using Microsoft.AspNetCore.Http;
 #endif
 using NLog.LayoutRenderers;
-using System.Collections.Generic;
-using NLog.Config;
 using NLog.Web.Enums;
-using System;
-using System.Linq;
 using NLog.Web.Internal;
 
 namespace NLog.Web.LayoutRenderers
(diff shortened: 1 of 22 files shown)


snakefoot:
It would actually be really nice if it didn't have to allocate / unwrap the HttpContextWrapper two times in AspNetLayoutRendererBase.

It performs a nullcheck and then the specialized version has to allocate / unwrap the HttpContextWrapper  again.

NLog.Web.AspNetCore/LayoutRenderers/AspNetRequestQueryStringLayoutRenderer.cs
```diff
@@ -41,12 +40,15 @@ public class AspNetQueryStringLayoutRenderer : AspNetLayoutMultiValueRendererBas
         /// <param name="logEvent"></param>
         protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
         {
-            var httpRequest = HttpContextAccessor?.HttpContext?.TryGetRequest();
+            var httpRequest = HttpContextAccessor.HttpContext.TryGetRequest();
             if (httpRequest == null)
+            {
+                Common.InternalLogger.Debug("aspnet-request-querystring - HttpContext Request is null");
```
304NotModified: this is now heavily duplicated. Can't be do this in the base class?

304NotModified:
What happend with all these codecov rerports?

snakefoot:
> What happend with all these codecov rerports?

No idea. Maybe it was super sensitive while I was working on it :)

NLog.Web.AspNetCore/LayoutRenderers/AspNetRequestQueryStringLayoutRenderer.cs
snakefoot: I guess it should be possible. Actually it could be nice if there was only one message for a single logevent. Right now one will get a debug-message for every triggered layout-renderer.

304NotModified: I can live with the duplicate message, but not really with the duplicate code ;)

snakefoot: Would it be okay if I just remove the debug-messages then. Because I find it difficult to modify these layoutrenderers without it being breaking interface changes.

snakefoot: And they are gone :)

PR #325, opened 2018-10-27
${AspNetRequestIp} added CheckForwardedForHeader to check for X-Forwarded-For header

Giorgi:


NLog.Web.AspNetCore/LayoutRenderers/AspNetRequestIpLayoutRenderer.cs
```diff
@@ -16,18 +19,52 @@ namespace NLog.Web.LayoutRenderers
...
         /// </summary>
         protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
         {
+            var forwardedIp = string.Empty;
             var httpContext = HttpContextAccessor.HttpContext;
+
 #if !ASP_NET_CORE
+            if (CheckForwardedForHeader)
+            {
+                var forwardedHeader = httpContext.TryGetRequest().Headers[ForwardedForHeader];
```
snakefoot: Instead of calling httpContext.TryGetRequest() twice, then do it once reuse it for `Headers[ForwardedForHeader]` and `ServerVariables["REMOTE_ADDR"]` (Maybe also handle null value from TryGetRequest)

Giorgi: Done

NLog.Web.AspNetCore/LayoutRenderers/AspNetRequestIpLayoutRenderer.cs
```diff
@@ -16,18 +19,59 @@ namespace NLog.Web.LayoutRenderers
...
+                    }
+                }
+            }
+
+            var ip = request.ServerVariables["REMOTE_ADDR"];
 #else
-            var ip = httpContext.Connection?.RemoteIpAddress;
+            if (CheckForwardedForHeader)
+            {
+                var forwardedHeaders = httpContext.Request.Headers.GetCommaSeparatedValues(ForwardedForHeader);
```
snakefoot: Maybe perform `if (Headers.ContainsKey(ForwardedForHeader))` before doing the `GetCommaSeparatedValues`-call. To reduce allocation when nothing to parse.

snakefoot:
Maybe possible to create a unit-test similar to `AspNetRequestReferrerRendererTests` where Headers are mocked?

NLog.Web.AspNetCore/LayoutRenderers/AspNetRequestIpLayoutRenderer.cs
Giorgi: If the header is missing `GetCommaSeparatedValues` returns `StringValues.Empty` which is a static field so I think it shouldn't cause any extra allocation

snakefoot: Not when reading the code:

https://github.com/aspnet/HttpAbstractions/blob/master/src/Microsoft.AspNetCore.Http.Abstractions/Extensions/HeaderDictionaryExtensions.cs#L42

Giorgi: OK, I will add that check

Giorgi: Done

Giorgi:
Tests added

snakefoot:
Thank you for the unit-tests. One last request from me. Could the logic be refactored so it becomes:

```c#
string ip = CheckForwardedForHeader ? TryLookupForwardHeader(httpContext) : string.Empty;
if (string.IsNullOrEmpty(ip))
{
#if !ASP_NET_CORE
      ip = httpContext.TryGetRequest()?.ServerVariables["REMOTE_ADDR"];
#else
      ip = httpContext.Connection?.RemoteIpAddress;
#endif
}
builder.Append(ip);
```

Then it would make the code more readable. Will also skip double lookup of ip-address. And maybe make the "Better Code Hub" more happy.

Giorgi:
What should be the type of context parameter in `TryLookupForwardHeader` method?

snakefoot:
Maybe something like this

```
#if !ASP_NET_CORE
string TryLookupForwardHeader(System.Web.HttpContext httpContext)
{
    // Blah Blah
}
#else
string TryLookupForwardHeader(Microsoft.AspNetCore.Http.HttpContext httpContext)
{
    // Blah Blah
}
#endif
```

Giorgi:
Do you think it's worth the hassle? I will have to call `httpContext.TryGetRequest()` twice again.

snakefoot:
Alternative like this:

```c#
{
   var httpRequest = httpContext.TryGetRequest();
   if (httpRequest==null)
      return;

   string ip = CheckForwardedForHeader ? TryLookupForwardHeader(httpRequest) : string.Empty;
   if (string.IsNullOrEmpty(ip))
   {
#if !ASP_NET_CORE
      ip = httpRequest?.ServerVariables["REMOTE_ADDR"];
#else
      ip = httpContext.Connection?.RemoteIpAddress;
#endif
   }
   builder.Append(ip);
}

#if !ASP_NET_CORE
string TryLookupForwardHeader(System.Web.HttpRequestBase httpRequest)
{
    // Blah Blah
}
#else
string TryLookupForwardHeader(Microsoft.AspNetCore.Http.HttpRequest httpRequest)
{
    var httpContext = httpRequest.HttpContext;
    // Blah Blah
}
#endif
```

Giorgi:
OK, I don't fully agree with that but I can do it.

snakefoot:
Thank you. All lights have become green. Now just need the final approval stamp from @304NotModified

304NotModified:
Thanks for the contribution! Also thanks @snakefoot for the reviews! 👍 🎉

PR #405, opened 2019-05-05
Fix AppVeyor build for NLog.Web on master branch

snakefoot:
VersionSuffix should also apply for NLog.Web-build

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,6 +33,9 @@ build_script:
     {
         $versionPrefix = $versionBuild
         $versionSuffix = "PR" + $env:APPVEYOR_PULL_REQUEST_NUMBER
+    }
+    if ($versionSuffix)
+    {
         $versionNuget = $versionPrefix + "-" + $versionSuffix
     }
     $build_aspnet = "build_aspnet.bat", "-nuget_version=$versionNuget"


snakefoot:
Not sure if https://www.nuget.org/packages/NLog.Web/4.8.2 should be unlisted?

304NotModified:
THanks!

> Not sure if https://www.nuget.org/packages/NLog.Web/4.8.2 should be unlisted?

We don't broke anything, so it's not really a problem. Leave it like that, as we can't create another 4.8.2 after all.

snakefoot:
But it looks like the PR worked. Now we also have:

https://www.nuget.org/packages/NLog.Web/4.8.2-rc1

But strange symbols has not been broken for:

https://www.nuget.org/packages/NLog.Web.AspNetCore/4.8.2-rc1

304NotModified:
They only break if there are code changes with the same version ;)

304NotModified:
> But it looks like the PR worked. Now we also have:
> 
> https://www.nuget.org/packages/NLog.Web/4.8.2-rc1

I think I will release the RTM for ASPnetcore soon, so we can focus on other stuff :)

snakefoot:
@304NotModified Thought you wanted to give the `${AspNetRequestPostedBody}` some more attention with splitting MaxContentLength into more parameters and adding ContentType-filter

304NotModified:
Yes, but I try to keep the aspnet4 and aspnetcore a bit in sync, so therefor releasing 4.8.2 now and we always add those features to another version

304NotModified:
I think it at least a very good improvement now :)

Issue #435, opened 2019-06-13
Is there a way to get rid of Object Reference exception for ${aspnet-sessionid} in internal logs for .net core 2. In case logger is being called in layer without HttpContext in it


Defee:
Hi All! 

## Issue Template Info:

**Type** (choose one):
**- Question**
My question is there a way to write when clause or other clause in layout to exclude the exception throwing from internal-logs in case HttpContext is not determined in the layer? 

It extremely hardens troubleshooting from internal nlog logs.

**NLog version**:  Resolved as 4.6.3

**NLog.Web/NLog.Web.AspNetCore version**: 
` <PackageReference Include="NLog.Web.AspNetCore" Version="4.8.2" />`

**NLog.Extensions.Logging version**: 
Resolved as 1.5.0 from NLog.Web.AspNetCore.

**Platform**: _.NET Core 2_ 

**Current NLog config** (xml or C#, if relevant)

```xml
<configuration>
  <configSection>
    <section name="nlog" type="NLog.NLogConfig.ConfigSectionHandler,NLog"/>
  </configSection>
  <nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" autoReload="true" internalLogLevel="info" internalLogFile="internal-nlog.txt">
    <extensions>
      <add assemblyFile="Microsoft.ApplicationInsights.NLogTarget.dll" />
    </extensions>
    <targets async="true">
      
      <target name="dbTarget" xsi:type="Database" 
              connectionString="YourConnectionString" 
              keepConnection="true" 
              commandText="YourSchema.SaveLogEvent 
                @timeStamp,
                @level, 
                @username, 
                @logger, 
                @url, 
                @machineName,
                @threadId,
                @sessionId,
                @referrer,
                @userAgent,
                @code,
                @message">
        <parameter name="@timeStamp" layout="${date}" />
        <parameter name="@level" layout="${level}" />
        <parameter name="@username" layout="${aspnet-user-identity}" />
        <parameter name="@logger" layout="${logger}" />
        <parameter name="@url" layout="${aspnet-request-url:IncludeHost=true:IncludePort=true:IncludeQueryString=true:IncludeScheme=true}" />
        <parameter name="@machineName" layout="${machinename}" />
        <parameter name="@threadId" layout="${threadid}" />
        <parameter name="@sessionId" layout="${aspnet-sessionid}" />
[...]

PR #436, opened 2019-06-13
aspnet-session - Reduce log-level to debug when session is not available

snakefoot:

PR #440, opened 2019-06-16
Fix for the Session Feature not configured exception

Defee:
Related issues: #435 

Hotfix for the Invalid Operation Exception when using sessionId in the middle layer where session is not available.

I think it might be also improved with checking the HttpContextAccessor is not not null, just in case :)

### DIFF
--- a/src/Shared/LayoutRenderers/AspNetSessionIdLayoutRenderer.cs
+++ b/src/Shared/LayoutRenderers/AspNetSessionIdLayoutRenderer.cs
@@ -1,4 +1,7 @@
 using System.Text;
+#if ASP_NET_CORE
+using Microsoft.AspNetCore.Http.Features;
+#endif
 using NLog.Common;
 using NLog.Config;
 using NLog.LayoutRenderers;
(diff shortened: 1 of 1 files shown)


src/Shared/LayoutRenderers/AspNetSessionIdLayoutRenderer.cs
```diff
@@ -23,8 +26,12 @@ public class AspNetSessionIdLayoutRenderer : AspNetLayoutRendererBase
         /// <param name="logEvent">Logging event.</param>
         protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
         {
+            var isSessionConfigured = true;
             var context = HttpContextAccessor.HttpContext;
-            if (context?.Session == null)
+#if ASP_NET_CORE
+            isSessionConfigured = context.Features.Get<ISessionFeature>() != null;
```
304NotModified: I prefer a more specific log message and return here (so we won't need the variable)

304NotModified:
FYI, I think this will conflict with https://github.com/NLog/NLog.Web/pull/436

(Not checked in depth, only on mobile now)

Defee:
> FYI, I think this will conflict with #436
> 
> (Not checked in depth, only on mobile now)

I think it will conflict. I noticed something when debugging. It is actually throwing the exception inside the "_Session_" property as per my understanding, AspNet Core checks the session configuration on the property call and throws InvalidOperation if no configuration found. 

So that's why it might be worth just to add the extra check for the session feature in the http context and let the other possible InvalidOperations flow ( just in case). But that's just a suggestion.

snakefoot:
Have updated #436 with the changes from this.

304NotModified:
Thanks, so superseded by #436 and that one is merged :)

PR #472, opened 2019-09-18
AspNetRequestCookieLayoutRenderer - Added support for Exclude

snakefoot:
Updated Microsoft.AspNetCore dependencies to Long Term Support versions

See also: https://dotnet.microsoft.com/platform/support/policy/dotnet-core

### DIFF
--- a/src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs
+++ b/src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs
@@ -73,19 +73,19 @@ namespace NLog.Web.LayoutRenderers
             var cookies = httpRequest.Cookies;
             if (cookies?.Count > 0)
             {
-                bool excludeKeys = (CookieNames == null || CookieNames.Count == 0) && Exclude?.Count > 0;
-                var cookieValues = GetCookieValues(cookies, excludeKeys);
+                bool checkForExclude = (CookieNames == null || CookieNames.Count == 0) && Exclude?.Count > 0;
+                var cookieValues = GetCookieValues(cookies, checkForExclude);
                 SerializePairs(cookieValues, builder, logEvent);
             }
         }
 
 #if !ASP_NET_CORE
-        private IEnumerable<KeyValuePair<string, string>> GetCookieValues(HttpCookieCollection cookies, bool excludeKeys)
+        private IEnumerable<KeyValuePair<string, string>> GetCookieValues(HttpCookieCollection cookies, bool checkForExclude)
         {
             var cookieNames = CookieNames?.Count > 0 ? CookieNames : cookies.Keys.Cast<string>().ToList();
             foreach (var cookieName in cookieNames)
             {
-                if (excludeKeys && Exclude.Contains(cookieName))
+                if (checkForExclude && Exclude.Contains(cookieName))
                     continue;
 
                 var httpCookie = cookies[cookieName];
@@ -116,12 +116,12 @@ namespace NLog.Web.LayoutRenderers
             }
         }
 #else
-        private IEnumerable<KeyValuePair<string, string>> GetCookieValues(IRequestCookieCollection cookies, bool excludeKeys)
+        private IEnumerable<KeyValuePair<string, string>> GetCookieValues(IRequestCookieCollection cookies, bool checkForExclude)
         {
             var cookieNames = CookieNames?.Count > 0 ? CookieNames : cookies.Keys;
             foreach (var cookieName in cookieNames)
             {
-                if (excludeKeys && Exclude.Contains(cookieName))
+                if (checkForExclude && Exclude.Contains(cookieName))
                     continue;
 
                 if (!cookies.TryGetValue(cookieName, out var cookieValue))
--- a/src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs
+++ b/src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs
@@ -25,6 +25,8 @@ namespace NLog.Web.LayoutRenderers
     /// <code lang="NLog Layout Renderer">
     /// ${aspnet-request-cookie:OutputFormat=Flat}
     /// ${aspnet-request-cookie:OutputFormat=Json}
+    /// ${aspnet-request-cookie:OutputFormat=Json:CookieNames=username}
+    /// ${aspnet-request-cookie:OutputFormat=Json:Exclude=access_token}
     /// </code>
     /// </example>
     [LayoutRenderer("aspnet-request-cookie")]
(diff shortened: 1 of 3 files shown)


304NotModified:
Thanks for the PR! I'm currently on holiday but will try to check this soon :)

src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs
```diff
@@ -37,6 +39,24 @@ public class AspNetRequestCookieLayoutRenderer : AspNetLayoutMultiValueRendererB
...
+#else
+        public HashSet<string> Exclude { get; set; }
+#endif
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AspNetRequestCookieLayoutRenderer" /> class.
+        /// </summary>
+        public AspNetRequestCookieLayoutRenderer()
+        {
+            Exclude = new HashSet<string>(new[] { "AUTH", "SESS_ID" }, StringComparer.OrdinalIgnoreCase);
```
304NotModified: I like the StringComparer.OrdinalIgnoreCase, but I'm not sure if that works when exclude options are set from config :/

src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
```diff
@@ -69,9 +69,9 @@ Supported platforms:
     <PackageReference Include="Microsoft.AspNetCore.Routing.Abstractions" Version="1.1.2" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'net461' ">
-    <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.2.0" />
-    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.2.0" />
-    <PackageReference Include="Microsoft.AspNetCore.Routing.Abstractions" Version="2.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.1.1" />
+    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.1.1" />
+    <PackageReference Include="Microsoft.AspNetCore.Routing.Abstractions" Version="2.1.1" />
```
304NotModified: Maybe we should state this in the readme (and/or add a comment here) so that others (including me) don't forget it ;)

src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs
```diff
@@ -111,10 +116,14 @@ private IEnumerable<string> GetCookieNames(Cookies cookies)
             }
         }
 #else
-        private IEnumerable<KeyValuePair<string, string>> GetCookies(IRequestCookieCollection cookies, IEnumerable<string> cookieNames)
+        private IEnumerable<KeyValuePair<string, string>> GetCookieValues(IRequestCookieCollection cookies, bool excludeKeys)
```
304NotModified: Minor, I prefer the same terms (names instead of keys)

src/Shared/LayoutRenderers/AspNetRequestFormLayoutRenderer.cs
```diff
@@ -48,12 +48,12 @@ public class AspNetRequestFormLayoutRenderer : AspNetLayoutMultiValueRendererBas
 #endif
 
         /// <summary>
-        /// Constructor
+        /// Initializes a new instance of the <see cref="AspNetRequestFormLayoutRenderer" /> class.
```
304NotModified: While this is indeed better, it has the same informational value. ;)

src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs
304NotModified: That's probably an existing issue then.

snakefoot: I guess changes could be make to fix this. But will make that a separate PR.

src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs
snakefoot: Think your mobile phone is cheating. I have moved the lookup of `cookieNames `into the `GetCookieValues`-method. It is a new parameter called `excludeKeys`

src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs
304NotModified: 👍

304NotModified: -> https://github.com/NLog/NLog/issues/3591

src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs
304NotModified: We have now `CookieNames` and `excludeKeys`. It should be `excludeNames` IMO (we can't change it to CookieKeys as that's a breaking change)

snakefoot:
Think you are still not seeing the source code correctly but cheated by your mobile


Sent from my Sony Xperia


---- Julian Verdurmen wrote ----


@304NotModified commented on this pull request.

In src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs<https://github.com/NLog/NLog.Web/pull/472#discussion_r326598190>:

> @@ -111,10 +116,14 @@ private IEnumerable<string> GetCookieNames(Cookies cookies)
             }
         }
 #else
-        private IEnumerable<KeyValuePair<string, string>> GetCookies(IRequestCookieCollection cookies, IEnumerable<string> cookieNames)
+        private IEnumerable<KeyValuePair<string, string>> GetCookieValues(IRequestCookieCollection cookies, bool excludeKeys)


We have now CookieNames and excludeKeys. It should be excludeNames IMO (we can't change it to CookieKeys as that's a breaking change)

src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs
snakefoot: Think you are still not seeing the source code correctly but cheated by your mobile

304NotModified: Don't you see the CookieNames property? And the excludeKeys parameter name?

snakefoot: Yes and cookieNames is now a variable inside the function (it is not gone, but NOT passed as parameter), and `excludeKeys` is passed as parameter to reuse the check for both NLog.Web and Nlog.Web.AspNetCore. Again you are not seeing the code, but just the diff on your phone.

304NotModified: No it not gone. My point is that name=key. So 2 terms for the same thing.

304NotModified: So excludeNames or excludeCookieNames

snakefoot: Ok renamed from `excludeKeys` to `checkForExclude`. No names. No keys.

snakefoot: I was confused by your sentence "we can't change it to CookieKeys as that's a breaking change".

src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
snakefoot: Well not the master of the universe. But usually sub-components at the bottom of the pyramid should be the slowest at updating to newer versions. Not a big fan for dependency-update-robots :)

src/Shared/LayoutRenderers/AspNetRequestCookieLayoutRenderer.cs
304NotModified: When I now see it, I guess I was also confused ;)

Anyway, perfect now! Thanks!

src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
304NotModified: I've created a new PR so this is more explicit/transparent. https://github.com/NLog/NLog.Web/pull/476

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/AspNetRequest-Cookie-Layout-Renderer

PR #459, opened 2019-09-05
Fixed double logging mixing UseNLog with AddNLog

snakefoot:
See also https://github.com/NLog/NLog.Extensions.Logging/pull/325

Requires release of NLog.Extension.Logging 1.5.4

PR #482, opened 2019-09-24
Removed obsolete-attribute from ConfigureNLog-methods for ILoggingBuilder

snakefoot:
Reverting #257. Since it will no longer generate double logging after #459 

Also fixed `ILoggingBuilder.ConfigureNLog()`-methods so they can work alone without `IHostBuilder.UseNLog()`.

See also: https://github.com/aspnet/AspNetCore.Docs/issues/14554

### DIFF
--- a/src/NLog.Web.AspNetCore/AspNetExtensions.cs
+++ b/src/NLog.Web.AspNetCore/AspNetExtensions.cs
@@ -80,11 +80,9 @@ namespace NLog.Web
         /// <param name="configFileName">Path to NLog configuration file, e.g. nlog.config. </param>
         /// >
         /// <returns>LogFactory to get loggers, add events etc</returns>
-        [Obsolete("Use UseNLog() on IWebHostBuilder, and NLog.Web.NLogBuilder.ConfigureNLog()")]
         public static LogFactory ConfigureNLog(this ILoggingBuilder builder, string configFileName)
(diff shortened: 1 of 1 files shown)


304NotModified:
I'm doubting about this. Afaik it's still bad practice to use both? Nothing goes wrong but it's still obsolete IMO

snakefoot:
The ILoggerBuilder interface is not obsolete. And you are not required to use IHostBuilder.

snakefoot:
But it would be nice to have a better fluent-interface-integration with ILoggerBuilder and IHostBuilder to load the NLog-config-file depending on environment (Ex. `NLog.Dev.config` and `NLog.Production.config`)

snakefoot:
I believe it should be possible to setup NLog as LoggingProvider directly on [LoggerFactory](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/index?view=aspnetcore-3.0#add-providers):

```
var loggerFactory = LoggerFactory.Create(builder =>
{
    builder.ConfigureNLog("NLog.config");
});
ILogger logger = loggerFactory.CreateLogger<Program>();
logger.LogInformation("Example log message");
```

This the recommended way when not using IHostBuilder.

304NotModified:
Ok merged thanks!

PR #484, opened 2019-09-27
Let IHostBuilder AddNLog work correctly with ${configsetting}

snakefoot:
IHostBuilder AddNLog attempt to setup ConfigSettingLayoutRenderer early

Increase the chance that fluent configuration will work.

### DIFF
--- a/src/NLog.Web.AspNetCore/AspNetExtensions.cs
+++ b/src/NLog.Web.AspNetCore/AspNetExtensions.cs
@@ -112,8 +112,13 @@ namespace NLog.Web
         /// <param name="configFileName">Path to NLog configuration file, e.g. nlog.config. </param>
         public static ILoggingBuilder AddNLog(this ILoggingBuilder builder, string configFileName)
         {
-            ConfigureServicesNLog(null, builder.Services, serviceProvider => serviceProvider.GetService<IConfiguration>());
-            LogManager.LoadConfiguration(configFileName);
+            AddNLogLoggerProvider(builder.Services, null, null, (serviceProvider, config, options) =>
+            {
+                var provider = CreateNLogLoggerProvider(serviceProvider, config, options);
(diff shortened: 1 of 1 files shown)


src/NLog.Web.AspNetCore/AspNetExtensions.cs
```diff
@@ -112,8 +112,11 @@ public static LogFactory ConfigureNLog(this ILoggingBuilder builder, LoggingConf
         /// <param name="configFileName">Path to NLog configuration file, e.g. nlog.config. </param>
         public static ILoggingBuilder AddNLog(this ILoggingBuilder builder, string configFileName)
         {
-            ConfigureServicesNLog(null, builder.Services, serviceProvider => serviceProvider.GetService<IConfiguration>());
-            LogManager.LoadConfiguration(configFileName);
+            ConfigureServicesNLog(null, builder.Services, (s, c) =>
```
304NotModified: Please use longer names then `s` and `c`. This isn't C ;)

304NotModified:
> Increase the chance that fluent configuration will work.

Could you elaborate this a bit? Not sure what we are to improve here.

snakefoot:
> Could you elaborate this a bit? Not sure what we are to improve here.

My initial version of AddNLog loaded the NLog-config before `IHostBuilder.Build()` had been called. Thus the registration code for setting up config (etc.) had not been called. Causing all Layout-renderers that depends on `${configsetting}` to fail.

By delaying the loading of NLog-config until `IHostBuilder.Build()` is called will help on this.

src/NLog.Web.AspNetCore/AspNetExtensions.cs
snakefoot: Found a better way that got rid of short names.

304NotModified:
clear! Thanks!

PR #524, opened 2020-02-23
aspnet-traceidentifier should automatically check Activity.Current.Id for AspNetCore3

### LINKED ISSUE #523: Wrong trace identifier on ASP.Net Core 3.0
I'm using the ${aspnet-traceidentifier} layout renderer to put the trace identifier in the logs, so I can match it with the value from the problem details report built into ASP.Net Core.

But the value from the problem details and from ${aspnet-traceidentifier} is now different.

ASP.Net Core is now using ActivityId.

Ref.:
https://github.com/dotnet/aspnetcore/blob/2e4274cb67c049055e321c18cc9e64562da52dcf/src/Mvc/Mvc.Core/src/Infrastructure/DefaultProblemDetailsFactory.cs#L90 

Ref.:
https://github.com/dotnet/aspnetcore/issues/18876#issuecomment-583484818

Is it possible to either change the renderer, or create a new one that does the same as asp.net core?

snakefoot:
This is what's used in ASP .NET Core 3 templates showing Request Id. Should do the same for consistency.

Resolves #523

See also: https://github.com/dotnet/aspnetcore/blob/2e4274cb67c049055e321c18cc9e64562da52dcf/src/Mvc/Mvc.Core/src/Infrastructure/DefaultProblemDetailsFactory.cs#L90

### DIFF
--- a/src/Shared/LayoutRenderers/AspNetTraceIdentifierLayoutRenderer.cs
+++ b/src/Shared/LayoutRenderers/AspNetTraceIdentifierLayoutRenderer.cs
@@ -4,7 +4,6 @@ using NLog.Config;
 using NLog.LayoutRenderers;
 #if ASP_NET_CORE
 using Microsoft.AspNetCore.Http;
-
 #endif
 
 namespace NLog.Web.LayoutRenderers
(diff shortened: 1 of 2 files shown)


src/Shared/LayoutRenderers/AspNetTraceIdentifierLayoutRenderer.cs
```diff
@@ -23,7 +22,20 @@ protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
             builder.Append(LookupTraceIdentifier(httpContext));
         }
 
-#if ASP_NET_CORE
+        /// <summary>
+        /// Ignore the System.Diagnostics.Activity.Current.Id value (AspNetCore3 uses ActivityId by default)
+        /// </summary>
+        public bool IgnoreActivityId { get; set; }
```
304NotModified: What about PreferActivityId (and then true by default?)

snakefoot: I like to keep the "action" words to a minimum for consistency (Enable / Disable / Ignore). And when something is enabled by default then I prefer the setting to have opposite name.

304NotModified: I prefer giving flags never negative names (so that you have false+false=true), but OK, it's not that important for here.

304NotModified:
maybe 4.10 is better for this?

Unfortunately NLog.Web (so not aspnetcore), has no changes after 4.9 (im trying to keep them in sync)

PR #540, opened 2020-03-30
LogManager.Setup() - LoadConfigurationFromAppSettings

snakefoot:
Now you can do this in NetCore3:

```c#
var logger = NLog.LogManager.Setup().LoadConfigurationFromAppSettings().GetCurrentClassLogger();
```

And it replaces this for NetCore3 (Could consider making it obsolete):

```c#
var logger = NLog.Web.NLogBuilder.ConfigureNLog("nlog.config").GetCurrentClassLogger();
```

And also configures ${configsetting} and also checks appsetting.json if it contains a NLog-section with NLog config. But if none found, then NLog automatic load of NLog.config takes over.

For NetCore2 then one can do this instead of `NLogBuilder.ConfigureNLog`:
```c#
var builder = new ConfigurationBuilder()
                .SetBasePath(basePath ?? Directory.GetCurrentDirectory())
                .AddJsonFile("appsettings.json", optional: false, reloadOnChange: true)
                .AddJsonFile($"appsettings.{environment}.json", optional: true)
                .AddEnvironmentVariables();
var config = builder.Build();
var logger = LogManager.Setup()
                       .RegisterNLogWeb(config)
                       .LoadConfigurationFromFile("nlog.config")
                       .GetCurrentClassLogger();
```

This PR is a proof-of-concept, and should be delayed until NLog.Extension.Logging has been released with NLog 4.7 (Or maybe NLog 4.7.1)

### DIFF
--- a/examples/ASP.NET Core 2/Visual Studio 2017/ASP.NET Core 2 - VS2017/Program.cs
+++ b/examples/ASP.NET Core 2/Visual Studio 2017/ASP.NET Core 2 - VS2017/Program.cs
@@ -3,6 +3,7 @@ using Microsoft.AspNetCore;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
+using NLog.Extensions.Logging;
 using NLog.Web;
 
 namespace NLog.Web.AspNetCore2.Example
@@ -11,7 +12,13 @@ namespace NLog.Web.AspNetCore2.Example
     {
         public static void Main(string[] args)
         {
-            var logger = NLog.Web.NLogBuilder.ConfigureNLog("nlog.config").GetCurrentClassLogger();
+            var config = new ConfigurationBuilder().Build();
+
+            var logger = LogManager.Setup()
+                                   .RegisterNLogWeb(config)
+                                   .LoadConfigurationFromFile("nlog.config")
+                                   .GetCurrentClassLogger();
+
             try
             {
                 logger.Debug("init main");
--- a/examples/ASP.NET Core 3/ASP.NET Core 3 - VS2019/Program.cs
+++ b/examples/ASP.NET Core 3/ASP.NET Core 3 - VS2019/Program.cs
@@ -6,6 +6,7 @@ using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using NLog;
 using NLog.Web;
 
 namespace ASP.NET_Core_3___VS2019
@@ -14,7 +15,10 @@ namespace ASP.NET_Core_3___VS2019
     {
         public static void Main(string[] args)
         {
-            var logger = NLog.Web.NLogBuilder.ConfigureNLog("nlog.config").GetCurrentClassLogger();
+            var logger = LogManager.Setup()
+                                   .LoadConfigurationFromAppSettings()
+                                   .GetCurrentClassLogger();
+
             try
             {
                 logger.Debug("init main");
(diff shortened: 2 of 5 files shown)


snakefoot:
Could be tempted to move the configure-everything-method out so it is an extension method to `LogManger.Setup()` directly. So it calls Setup().SetupExtensions(s => s.Register(NLog.Web.AspNetCore)) instead of doing massage of ConfigurationItemFactory.Default.

Maybe the same thought should be adjusted for https://github.com/NLog/NLog.Extensions.Logging/pull/410 where loading of NLogLoggingConfiguration should also be placed as an extension method to `LogManger.Setup()` directly. And setting up configsetting-layoutrenderer should be extension method to `LogManger.Setup().SetupExtensions()`

snakefoot:
Notice this is proof-on-concept and should not be merged, since it will bump to NLog 4.7

NLog.Extensions.Logging ver. 1.6.3 should be released first. Maybe after NLog 4.7.1 has been released. And then NLog.Web.AspNetCore can update its dependency, and then this can be merged.

snakefoot:
I have my doubts about having two extension methods with the same name but different namespaces:
  - NLog.Web.LoadNLogConfigFromSection
  - NLog.Extensions.Logging.LoadNLogConfigFromSection

Both methods does the same, but NLog.Web has the additional logic that it automatically registers NLog.Web-extensions before loading the NLogConfig.

But I think confusion in tutorials may occur, and also confusion when right-clicking the method to fix missing usings (Will show two possible namespaces to include)

304NotModified:
Have to think about it. It could be confusing, maybe there is a better solution.

304NotModified:
What about LoadNLogConfigFromSectionWithWebIntegration? (so something like that?)

snakefoot:
> What about LoadNLogConfigFromSectionWithWebIntegration? (so something like that?)

Sounds like an awful name. Then prefer the name collission that also exists with AddNLog and UseNLog

snakefoot:
What about doing this in NetCore2 (Introducing `SetupNLogWeb()` and instead reuse `LoadNLogConfigFromSection` from NLog.Extension.Logging)

```
var logger = NLog.LogManager.Setup().SetupNLogWeb().LoadNLogConfigFromSection(config).LogFactory.GetCurrentClassLogger();
```

304NotModified:
> Sounds like an awful name. Then prefer the name collission that also exists with AddNLog and UseNLog

Agree, but that even better to me than a collision ;) (no joke).

304NotModified:
> What about doing this in NetCore2 (Introducing `SetupNLogWeb()` and instead reuse `LoadNLogConfigFromSection` from NLog.Extension.Logging)

Sounds good!

snakefoot:
> maybe stupid question, but we can't get rid if Setup() and .LogFactory? (it's a bit hard for an end-user I guess)

We can easily add a GetCurrentClassLogger()-extension-method to the ISetupBuilder-interface. 

Not sure what you mean about removing `Setup()`, can you show some code for implementing this?

304NotModified:
> Not sure what you mean about removing `Setup()`, can you show some code for implementing this?

I dislike 2 times setup, but with  AddNLogWeb  this is already fixed :)

snakefoot:
Still waiting for a NLog.Extension.Logging with dependency on NLog 4.7.2

304NotModified:
> Still waiting for a NLog.Extension.Logging with dependency on NLog 4.7.2 (This PR introduces direct reference to NLog because it is missing)

Master has been updated!

snakefoot:
@304NotModified Have added unit-tests for the new method `LoadConfigurationFromAppSettings()`, but it seems code-coverage is not seeing them.

304NotModified:
Thanks

I think that coverage is only for one platform (merge same code base for multiple platforms is always difficult, also in terms of performance)

304NotModified:
Is this ready to review? (Unexpected force pushes :))

snakefoot:
> Is this ready to review? (Unexpected force pushes :))

Yes ready now. Just wanted to add `AutoLoadAssemblies(false)` to `RegisterNLogWeb`

304NotModified:
Thanks! Merged!

PR #549, opened 2020-04-19
${aspnet-request-posted-body}: Use Async reading to prevent "Synchronous operations are disallowed"

### LINKED ISSUE #548: ${aspnet-request-posted-body} could throw 'Synchronous operations are disallowed' for ASP.NET Core 3
NLog 4.7.0
Nlog.Target.Http 1.0.1
Nlog.web.AspNet.Core 4.9.1
Aspnet.Core 3.1.201

internal logging first was indicating : 
```
Warn Exception in 'NLog.Web.LayoutRenderers.AspNetRequestPostedBody.Append()' 
Exception: System.InvalidOperationException: 
Synchronous operations are disallowed. 
Call ReadAsync or set AllowSynchronousIO to true instead.
```

After googling I found that I must add 

```c#
 services.Configure<IISServerOptions>(options =>
 {
     options.AllowSynchronousIO = true;
  });
````

_Originally posted by @husseinshaib1 in https://github.com/NLog/NLog/issues/3906#issuecomment-616199275_

### PR DESCRIPTION
**Pull request by 304NotModified**
needs a good (real-life) test 

fixes #548

### CODE REVIEW

File: src/Shared/LayoutRenderers/AspNetRequestPostedBody.cs [hunk 1]
```diff
@@ -70,7 +72,12 @@ private static string BodyToString(Stream body)
                 // Note: don't dispose the StreamReader, it will close the stream and that's unwanted. You could pass that that
                 // to the StreamReader in some platforms, but then the dispose will be a NOOP, so for platform compat just don't dispose
                 var bodyReader = new StreamReader(body);
+
+#if NETSTANDARD
+                var content = bodyReader.ReadToEndAsync().RunTaskSynchronously();
```
snakefoot, line 77: Would it not be better this: `bodyReader.ReadToEndAsync().ConfigureAwait(false).GetAwaiter().GetResult()` ?

304NotModified: I was under the impression that GetAwaiter().GetResult() could deadlock?

I don't do this stuff regularly, most of the time I use proper `awaits`

snakefoot: The secret sauce is the `ConfigureAwait(false)`

304NotModified: Not really clear answer, guess it's OK then:

https://stackoverflow.com/questions/39007006/is-getawaiter-getresult-safe-for-general-use

I will change to `ConfigureAwait(false).GetAwaiter().GetResult()`

304NotModified: Except we also support ASP.NET non-core? Maybe drop that change for ASP.NET non-core

snakefoot: Also think the async-logic should be dropped for ASP.NET non-core.

304NotModified: Fixed!

304NotModified:
Update the demo and confirm it works now:

image

snakefoot:
Thanks to @bakgerman now NLog.Web.AspNetCore v5.1 re-introduces [${aspnet-request-posted-body}](https://github.com/NLog/NLog/wiki/AspNet-Request-posted-body-layout-renderer) with help from middleware:
```c#
app.UseMiddleware<NLog.Web.NLogRequestPostedBodyMiddleware>();
```
It is no longer necessary to explicit call `context.Request.EnableBuffering();` as it is handled by the middleware.

PR #570, opened 2020-05-27
Handle int for session value

304NotModified:
see https://stackoverflow.com/questions/62054400/can-the-nlog-aspnet-sessionvariable-intid-renderer-extract-display-integers-f/62054553#62054553

I think the only solution here is to add a type option to the renderer. What do you think @snakefoot ?

(If it's 4 bytes long, we only know it **could** be an int)

PS: these are the extensions in ASP.NET Core:

```c#
namespace Microsoft.AspNetCore.Http
{
    public static class SessionExtensions
    {
        public static void SetInt32(this ISession session, string key, int value)
        {
            var bytes = new byte[]
            {
                (byte)(value >> 24),
                (byte)(0xFF & (value >> 16)),
                (byte)(0xFF & (value >> 8)),
                (byte)(0xFF & value)
            };
            session.Set(key, bytes);
        }

        public static int? GetInt32(this ISession session, string key)
        {
            var data = session.Get(key);
            if (data == null || data.Length < 4)
            {
                return null;
            }
            return data[0] << 24 | data[1] << 16 | data[2] << 8 | data[3];
        }

        public static void SetString(this ISession session, string key, string value)
        {
            session.Set(key, Encoding.UTF8.GetBytes(value));
        }

        public static string GetString(this ISession session, string key)
        {
            var data = session.Get(key);
            if (data == null)
            {
                return null;
            }
            return Encoding.UTF8.GetString(data);
        }

        public static byte[] Get(this ISession session, string key)
        {
            byte[] value = null;
            session.TryGetValue(key, out value);
            return value;
        }
    }
}
```

### DIFF
--- a/src/NLog.Web.AspNetCore/LayoutRenderers/SessionValueType.cs
+++ b/src/NLog.Web.AspNetCore/LayoutRenderers/SessionValueType.cs
@@ -1,0 +1,20 @@
+#if ASP_NET_CORE
+
+namespace NLog.Web.LayoutRenderers
+{
+    /// <summary>
+    /// The type of a value
+    /// </summary>
+    public enum SessionValueType
+    {
+        /// <summary>
+        /// String
+        /// </summary>
+        String,
+        /// <summary>
+        /// Int32
+        /// </summary>
+        Int32
+    }
+}
+#endif
--- a/src/Shared/Internal/HttpContextExtensions.cs
+++ b/src/Shared/Internal/HttpContextExtensions.cs
@@ -82,6 +82,20 @@ namespace NLog.Web.Internal
 
             return Encoding.UTF8.GetString(data);
         }
+
+        public static int? GetInt32(this ISession session, string key)
+        {
+            if (!session.TryGetValue(key, out var data))
+            {
+                return null;
+            }
+
+            if (data == null || data.Length < 4)
+            {
+                return null;
+            }
+            return data[0] << 24 | data[1] << 16 | data[2] << 8 | data[3];
+        }
 #endif
 
 #if !ASP_NET_CORE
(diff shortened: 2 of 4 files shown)


snakefoot:
No opinion. But maybe just accept that string-values is the only way.

snakefoot:
Strange that Microsoft have not added some kind of byte-order-mark to signal whether byte-array is an integer or string.

304NotModified:
Or just an int which would be utf8 compatible. Yes, less efficient. This int "encoding" is just unsafe stuff

snakefoot:
Yes also think it would be nicer that the SetInt32-extension-method, just converted integer into utf8-byte-array without doing string-allocation first. Not sure people using session-state-objects care much about 4 or 10 bytes.

src/NLog.Web.AspNetCore/LayoutRenderers/SessionValueType.cs
```diff
@@ -0,0 +1,20 @@
+#if ASP_NET_CORE
+
+namespace NLog.Web.LayoutRenderers
+{
+    /// <summary>
+    /// The type of a value
+    /// </summary>
+    public enum SessionValueType
```
snakefoot, line 8: Like the name

src/Shared/Internal/HttpContextExtensions.cs
```diff
@@ -62,7 +62,7 @@ internal static HttpResponse TryGetResponse(this HttpContext context)
         }
 #endif
 
-#if ASP_NET_CORE
+#if ASP_NET_CORE1 || ASP_NET_CORE2
```
snakefoot, line 65: Good idea since ASP_NET_CORE3 includes entire framework

304NotModified:
updated docs: https://github.com/NLog/NLog/wiki/AspNetSession-layout-renderer

PR #591, opened 2020-07-10
Added option to trim aspnet-request-posted-body.

Neophear:
With this change this:
`{_______"test":_"Some_____value____"_______}`
becomes this:
`{_"test":_"Some_value_"_}`
in the log.

Possible future change is to only trim outside json values.


304NotModified:
Thanks! But I'm not sure if this is needed, we have already https://github.com/NLog/NLog/wiki/Replace-Layout-Renderer and combining all possibilities isn't really maintainable

Neophear:
Ah I wasn't aware of that option. That makes sense!

snakefoot:
And it can be combined with https://github.com/NLog/NLog/wiki/Trim-Whitespace-Layout-Renderer

`${aspnet-request-posted-body:trimWhiteSpace=true}`

PR #602, opened 2020-08-19
Introduced AddNLogWeb to avoid conflict with AddNLog from NLog.Extension.Logging

snakefoot:
Reduce the level of confusion when used within `services.AddLogging()`-method. See also https://github.com/NLog/NLog/issues/4090

Like this:

```c#
            services.AddLogging(loggingBuilder =>
            {
                loggingBuilder.ClearProviders();
                loggingBuilder.AddNLogWeb(); // fix Missing serviceProvider, so no HttpContext
            });
```

Instead of this:

```c#
            services.AddLogging(loggingBuilder =>
            {
                loggingBuilder.ClearProviders();
                loggingBuilder.AddNLog(provider => null);  // fix Missing serviceProvider, so no HttpContext
            });
```

Guess when NLog.Extension.Logging ver. 2 with released with NLog ver. 5.0, then `AddNLog` can be marked as obsolete (Redirect to `AddNLogWeb`)

### DIFF
--- a/src/NLog.Web.AspNetCore/AspNetExtensions.cs
+++ b/src/NLog.Web.AspNetCore/AspNetExtensions.cs
@@ -111,11 +111,44 @@ namespace NLog.Web
         }
 
         /// <summary>
+        /// Enable NLog as logging provider for Microsoft Extension Logging
+        /// </summary>
+        /// <param name="builder">The logging builder</param>
+        public static ILoggingBuilder AddNLogWeb(this ILoggingBuilder builder)
+        {
+            return builder.AddNLogWeb(NLogAspNetCoreOptions.Default);
+        }
+
+        /// <summary>
+        /// Enable NLog as logging provider for Microsoft Extension Logging
+        /// </summary>
+        /// <param name="builder">The logging builder</param>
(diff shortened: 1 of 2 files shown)


snakefoot:
An alternative could be to also have `LoadConfigurationFromSection` in NLog.Web-namespace. Thus avoiding the need to have both `using NLog.Web` and `using NLog.Logging.Extensions` (Continue to have methods with name-conflicts)

304NotModified:
> An alternative could be to also have LoadConfigurationFromSection in NLog.Web-namespace. Thus avoiding the need to have both using NLog.Web and using NLog.Logging.Extensions

Related: https://github.com/NLog/NLog.Extensions.Logging/issues/421

304NotModified:
> An alternative could be to also have `LoadConfigurationFromSection` in NLog.Web-namespace. Thus avoiding the need to have both `using NLog.Web` and `using NLog.Logging.Extensions` (Continue to have methods with name-conflicts, but making it obvious that they have different features)

That's fine, in the end we will move everything to Microsoft.Extensions.DependencyInjection namespace, see NLog/NLog.Extensions.Logging#421. But that will be the next major version I guess

PR #614, opened 2020-12-12
ASPNETCORE_ENVIRONMENT takes now precedence over DOTNET_ENVIRONMENT

snakefoot:
See also: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-5.0#environments

image

### DIFF
--- a/src/NLog.Web.AspNetCore/Config/SetupBuilderExtensions.cs
+++ b/src/NLog.Web.AspNetCore/Config/SetupBuilderExtensions.cs
@@ -20,7 +20,7 @@ namespace NLog.Web
         /// </summary>
         public static ISetupBuilder LoadConfigurationFromAppSettings(this ISetupBuilder setupBuilder, string basePath = null, string environment = null, string nlogConfigSection = "NLog", bool optional = true, bool reloadOnChange = false)
         {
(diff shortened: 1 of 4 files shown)


snakefoot:
Maybe Azure Pipelines Config must be changed to use Nuget 5.8.0 ?

snakefoot:
Close and re-open to trigger rebuild

304NotModified:
I think this small behavior change needs a minor version bump

304NotModified:
It would be nice if we could unit test this. :)

snakefoot:
> It would be nice if we could unit test this. :)

Not sure how to perform mocking with extension-methods.

304NotModified:
A yes good one. You can't mock extensions (as it's syntactic sugar), the underlying calls should be mocked for that

snakefoot:
Guess one could make an internal extension-method that requires an interface-parameter which is used.


Sent from my Sony Xperia


---- Julian Verdurmen wrote ----


A yes good one. You can't mock extensions (as it's syntactic sugar), the underlying calls should be mocked for that

304NotModified (approved):

304NotModified:
I've created a new PR for the license, as we needed that also for other repos and thats easier to track (now and in the futher)

304NotModified:
you're right about the mocking. Environment.GetEnvironmentVariable can't be mocked easily, and if we do, we still can't assert something good as LoadConfigurationFromFile could not be tested easily.

To for now we keep it like this

PR #624, opened 2020-12-29
Version 4.10

304NotModified:


304NotModified:
Not really a large update for NLog.Web (non aspnet core). Doubting if we should keep the version in-sync

snakefoot:
> Not really an update for NLog.Web (non aspnet core). Doubting if we should keep the version in-sync. What do you think?

I think they should be kept in sync, and today NLog.Web is just a version-bump of NLog.

Issue #633, opened 2021-02-16
Could not find value for entry assembly and version type Assembly - Only when async=true


Giorgi:
I get `Could not find value for entry assembly and version type Assembly` but only if `async="true"` is specified.

**Type** (choose one):

- Bug

**NLog version**: 4.7.7

**NLog.Web version**: 4.10.0

**NLog.Extensions.Logging version**: 

**Platform**:  .Net 4.6.1

**Current NLog config** (xml or C#, if relevant)

```xml
<?xml version="1.0" encoding="utf-8" ?>
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">

  <targets async="true">
    <target name="db" xsi:type="Database" connectionStringName="PRODUCTION">
      <commandText>
        INSERT INTO [dbo].[ApplicationLog] ([LogLevel], [Logger], [UserId], [Message], [ErrorMessage], [ProjectName], [Url],
        [ServerIp], [ServerName],[RequestIp], [ProcessId], [ThreadId], [Properties], [AppVersion], [RequestId])
        VALUES (@LogLevel, @Logger, @UserId, @Message, @ErrorMessage, @ProjectName, @Url, @ServerIp, @ServerName, @RequestIp, @ProcessId, @ThreadId, @Properties, @AppVersion, @RequestId)</commandText>
      <parameter name="@LogLevel" layout="${level}" dbType="SqlDbType.Varchar" />
      <parameter name="@Logger" layout="${logger}" dbType="SqlDbType.Varchar" />
      <parameter name="@UserId" layout="${mdlc:item=UserId}" dbType="SqlDbType.Varchar"  />
      <parameter name="@Message" layout="${message}" dbType="SqlDbType.VarChar" />
      <parameter name="@ErrorMessage" layout="${exception:format=ToString:maxInnerExceptionLevel=5:innerFormat=ToString}" dbType="SqlDbType.VarChar"  />
      <parameter name="@ProjectName" layout="${gdc:item=Project}" dbType="SqlDbType.VarChar" />
      <parameter name="@Url" layout="${aspnet-request-url:IncludeQueryString=true}" dbType="SqlDbType.VarChar"  />
      <parameter name="@ServerIp" layout="${local-ip:cachedSeconds=60}" dbType="SqlDbType.VarChar"  />
[...]

PR #326, opened 2018-10-28
Added ThreadSafe-attribute for LayoutRenderer to optimize async Precalculate

snakefoot:
When using AsyncWrapper then logger-threads will be able to precalculate-layouts concurrently without waiting for Target-lock.

- ${aspnet-request-url} now renders directly to StringBuilder
- ${aspnet-request} now also checks HasFormContentType. See also #322 (Also enabled unit tests for NetCore)

PR #634, opened 2021-02-16
Remove ThreadAgnostic from AssemblyVersionLayoutRenderer when ASP.NET

snakefoot:
Trying to fix bug #633 on ASP.NET that was introduced with #305 (But not made better by #326)

Special hack implemented, since class-attributes are inherited and cannot be removed again.

Added unit-test to prove the fix is working.


snakefoot:
Seems the bug was initially introduced with #305 (but it did not become better with #326)

PR #639, opened 2021-02-26
Enable deterministic build

304NotModified:
See https://github.com/clairernovotny/DeterministicBuilds

and https://github.com/NLog/NLog.Extensions.Logging/pull/479

PR #647, opened 2021-02-28
Move to build.ps1

304NotModified:
More in sync with https://github.com/NLog/NLog.Extensions.Logging


snakefoot:
Love the idea of getting of the bat-file. 

But curious how to check if ContinuousIntegrationBuild is working. Tried opening the artifacts, and they showed red flags when viewing the nuget-packages.

snakefoot (approved):

304NotModified:
> But curious how to check if ContinuousIntegrationBuild is working. Tried opening the artifacts, and they showed red flags when viewing the nuget-packages

That only works with source link and pdb on NuGet. See https://github.com/NLog/NLog.Web/pull/639

PR #697, opened 2021-09-16
Introduced ValuesAsProperties to replace SingleAsArray

snakefoot:
Affects the following layout-renderers:

- AspNetRequestCookieLayoutRenderer
- AspNetRequestFormLayoutRenderer
- AspNetRequestHeadersLayoutRenderer
- AspNetQueryStringLayoutRenderer

I'm open for different property-name that gives better meaning.


304NotModified:
is it now impossible to render a single item as an array? Why is that? I don't see the rationale for this change, 



> Very tempted to introduce OutputFormat=JsonDictionary

That would be indeed great, also If we could have OutputFormat=JsonArray

snakefoot:
> is it now impossible to render a single item as an array? Why is that? I don't see the rationale for this change,

ValuesAsProperties = false means always as array, even when single-item.

304NotModified:
ah so a single item without array or object is now not possible, correct?

snakefoot:
> ah so a single item without array or object is now not possible?

Yes either you always get Json-Dictionary or always Json-Array. Not a mixture of both.

PR #700, opened 2021-09-26
Added W3CLoggerLayout for writing W3C Extended Logs

snakefoot:


snakefoot:
Further optimizations can be made when NLog 5.0-preview2 is out with:
- https://github.com/NLog/NLog/pull/4586 (Allow aspnet-request-duration work on NetCore)
- https://github.com/NLog/NLog/pull/4584 (Reduce memory allocations)

src/NLog.Web.AspNetCore/NLogRequestLoggingMiddleware.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// Executes the middleware.
+        /// </summary>
+        public async Task Invoke(HttpContext httpContext)
+        {
+            try
+            {
+                await _next(httpContext);
+                LogHttpRequest(httpContext, null);
+            }
+            catch (Exception exception) when (LogHttpRequest(httpContext, exception))
```
304NotModified, line 49: this is bit confusing, the `when` condition has a side effect?

304NotModified: is this intended?

src/NLog.Web.AspNetCore/NLogRequestLoggingMiddleware.cs
```diff
@@ -0,0 +1,77 @@
...
+                {
+                    _logger.LogWarning("HttpRequest Failed");
+                }
+                else
+                {
+                    _logger.LogInformation("HttpRequest Completed");
+                }
+            }
+
+            return false;
```
304NotModified, line 74: What is this return value saying?

src/NLog.Web/NLogRequestLoggingModule.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+            catch
+            {
+                // Nothing to do
+            }
+            finally
+            {
+                if (exception != null)
+                    Logger.Error(exception, "HttpRequest Exception");
+                else if (statusCode >= 400)
```
304NotModified, line 36: I guess we could also have  a 500+ without exception, I guess that should be an error

src/NLog.Web/NLogRequestLoggingModule.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        private void LogHttpRequest(object sender, EventArgs e)
+        {
+            Exception exception = null;
+            int statusCode = 400;
+
+            try
+            {
+                exception = HttpContext.Current?.Server?.GetLastError();
+                statusCode = HttpContext.Current?.Response?.StatusCode ?? 400;
```
304NotModified, line 26: I don't get this default of 400. Maybe we should make it at least configurable. or write below `statusCode >= 400 or statuscode is null` (pseudecode)

src/Shared/LayoutRenderers/AspNetRequestDurationLayoutRenderer.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// </summary>
+        /// <docgen category='Rendering Options' order='10' />
+        public System.Globalization.CultureInfo Culture { get; set; } = System.Globalization.CultureInfo.InvariantCulture;
+
+        /// <inheritdoc/>
+        protected override void InitializeLayoutRenderer()
+        {
+            if (DurationMsFormat == null && string.IsNullOrEmpty(Format) && ReferenceEquals(Culture, System.Globalization.CultureInfo.InvariantCulture))
+            {
+                System.Threading.Interlocked.CompareExchange(ref DurationMsFormat, System.Linq.Enumerable.ToArray(System.Linq.Enumerable.Select(System.Linq.Enumerable.Range(0, 1000), i => i.ToString())), null);
```
304NotModified, line 37: what are we trying to do in `System.Linq.Enumerable.ToArray(System.Linq.Enumerable.Select(System.Linq.Enumerable.Range(0, 1000), i => i.ToString())`? Maybe create a method for it as "documentation"

src/Shared/LayoutRenderers/AspNetRequestDurationLayoutRenderer.cs
```diff
@@ -0,0 +1,186 @@
...
+            }
+            else
+            {
+                return default(TimeSpan?);
+            }
+        }
+#endif
+
+#if ASP_NET_CORE && !ASP_NET_CORE2
+        private static TimeSpan? GetDuration(System.Diagnostics.Activity activity)
```
304NotModified, line 153: I'm not sure if all the paths has unit tests. I would recommend to add them

src/Shared/LayoutRenderers/AspNetRequestDurationLayoutRenderer.cs
```diff
@@ -0,0 +1,186 @@
...
+                return duration.ToString();
+            else
+                return string.Format(_formatString, duration);
+#else
+            return duration.ToString(Format, Culture);
+#endif
+        }
+
+#if !ASP_NET_CORE
+        private static TimeSpan? GetDuration(DateTime contextTimestamp)
```
304NotModified, line 134: I'm not sure if all the paths has unit tests. I would recommend to add them

src/Shared/LayoutRenderers/AspNetRequestDurationLayoutRenderer.cs
```diff
@@ -0,0 +1,186 @@
...
+                    builder.Append(".0");
+                }
+            }
+            else
+            {
+                builder.Append(durationMs.ToString("0.###", Culture));
+            }
+        }
+
+        string FormatDuration(TimeSpan duration)
```
304NotModified, line 121: I'm not sure if all the paths has unit tests. I would recommend to add them

src/Shared/LayoutRenderers/AspNetRequestDurationLayoutRenderer.cs
```diff
@@ -0,0 +1,186 @@
...
+                }
+                else
+                {
+                    builder.Append(FormatDuration(duration.Value));
+                }
+            }
+#endif
+        }
+
+        private void RenderDurationMs(StringBuilder builder, double durationMs)
```
304NotModified, line 79: I'm not sure if all the paths has unit tests. I would recommend to add them

src/NLog.Web/NLogRequestLoggingModule.cs
snakefoot: Not a ASP.NET developer, so clue when a HttpStatusCode should signal a warning. Just assumed anything above 400 was a serious error (and not just a redirection). But I can also see other definitions: https://github.com/microsoft/referencesource/blob/5697c29004a34d80acdaf5742d7e699022c64ecd/System/net/System/Net/Http/HttpResponseMessage.cs#L117

src/NLog.Web.AspNetCore/NLogRequestLoggingMiddleware.cs
snakefoot: Yes this is the new Exception-filter-logic, that allows one to react to the first-chance exception (keeping scope state). Instead of handling it in the normal exception-catch where one have left the scope. See also: https://docs.microsoft.com/en-us/dotnet/standard/exceptions/using-user-filtered-exception-handlers

src/NLog.Web.AspNetCore/NLogRequestLoggingMiddleware.cs
snakefoot: Says that the first-chance-exception-filter has not handled the exception, so it should just continue as a normal exception.

src/NLog.Web/NLogRequestLoggingModule.cs
snakefoot: Not a ASP.NET developer, so clue when a HttpStatusCode should signal a warning or error.

src/NLog.Web/NLogRequestLoggingModule.cs
snakefoot: Removed 400 as default and instead signal warning, when StatusCode < 100

src/Shared/LayoutRenderers/AspNetRequestDurationLayoutRenderer.cs
snakefoot: Think I will extract AspNetRequestDurationLayoutRenderer into its own pull-request. Where more attention can be given to test-coverage. I just wanted to see how easy it was to use NLog for request-logging.

src/NLog.Web.AspNetCore/NLogRequestLoggingMiddleware.cs
304NotModified: From my first view, I would say that this is the same:

```c#
catch (Exception exception){
LogHttpRequest(httpContext, exception)
}

So the difference is the scope? It would be great if we could ensure this from the tests.

Please document in code that we do this "trick" for the scope :)
```

src/Shared/LayoutRenderers/AspNetRequestDurationLayoutRenderer.cs
304NotModified: good idea!

snakefoot:
Created wiki-page: https://github.com/NLog/NLog/wiki/W3CExtendedLogLayout

PR #736, opened 2021-12-01
Added Support for ASP.NET Core 6 + Sample

ThomasArdal:
Sample for ASP.NET Core 6.0 on .NET 6. I had to call `Dispose` on the log factory in some of the unit tests to make them run. The problem is that the generated test file is locked when trying to read it. Another issue that I stumbled upon is that when targeting `net5.0` in `NLog.Web.AspNetCore` a constant named `ASP_NET_CORE3` is defined. We could consider changing this to `ASP_NET_CORE5` for .NET 5 and `ASP_NET_CORE6` for .NET 6. This will require some additional code changes, though.


ThomasArdal:
Part of this PR is also to review this new page: https://github.com/NLog/NLog/wiki/Getting-started-with-ASP.NET-Core-6

examples/ASP.NET Core 6/ASP.NET Core 6 NLog Example/Program.cs
```diff
@@ -7,6 +7,9 @@
 {
     var builder = WebApplication.CreateBuilder(args);
 
+    builder.Logging.ClearProviders();
+    builder.Logging.SetMinimumLevel(Microsoft.Extensions.Logging.LogLevel.Trace);
```
snakefoot, line 11: Think it would make sense if these lines was moved down, so it becomes:

```c#
// NLog: Setup NLog for Dependency injection
builder.Logging.ClearProviders();
builder.Logging.SetMinimumLevel(Microsoft.Extensions.Logging.LogLevel.Trace);
builder.Host.UseNLog();
```

P.S. With NLog 5.0 then `SetMinimumLevel` is no longer necessary. And if using `UseNLog(new NLogAspNetCoreOptions() { ReplaceLoggerFactory = true })` then both lines can be removed.

appveyor.yml
```diff
@@ -1,7 +1,7 @@
 version: 5.0.0.{build}
 clone_folder: c:\projects\nlogweb
 configuration: Release
-image: Visual Studio 2019
+image: Visual Studio 2022
```
snakefoot, line 4: Thank you for bumping. Should bump the `image` for NLog and NLog.Extensions.Logging before NLog 5.0 is released.

examples/ASP.NET Core 6/ASP.NET Core 6 NLog Example/nlog.config
```diff
@@ -0,0 +1,42 @@
...
+    <!-- File Target for all log messages with basic details -->
+    <target xsi:type="File" name="allfile" fileName="c:\temp\nlog-AspNetCore6-all-${shortdate}.log"
+            layout="${longdate}|${event-properties:item=EventId_Id:whenEmpty=0}|${uppercase:${level}}|${logger}|${message} ${exception:format=tostring}" />
+
+    <!-- File Target for own log messages with extra web details using some ASP.NET core renderers -->
+    <target xsi:type="File" name="ownFile-web" fileName="c:\temp\nlog-AspNetCore6-own-${shortdate}.log"
+            layout="${longdate}|${event-properties:item=EventId_Id:whenEmpty=0}|${uppercase:${level}}|${logger}|${message} ${exception:format=tostring}|url: ${aspnet-request-url}|action: ${aspnet-mvc-action}|${callsite}| body: ${aspnet-request-posted-body}" />
+
+    <!--Console Target for hosting lifetime messages to improve Docker / Visual Studio startup detection -->
+    <target xsi:type="Console" name="lifetimeConsole" layout="${level:truncate=4:lowercase=true}\: ${logger}[0]${newline}      ${message}${exception:format=tostring}" />
```
snakefoot, line 25: Could this be updated to match the wiki-page, so it becomes:
```xml
<target xsi:type="Console" name="lifetimeConsole" layout="${MicrosoftConsoleLayout}" />
```

examples/ASP.NET Core 6/ASP.NET Core 6 NLog Example/Program.cs
ThomasArdal: I wasn't sure about upgrading the docs to NLog 5. I agree, once released in stable it should target v5. What do you think? At least the package references should specify full version names if we target v5. Unless there's a way to create wildcards for prefixes that I don't know about. Currently it looks like this:

```xml
<ItemGroup>
  <PackageReference Include="NLog.Web.AspNetCore" Version="4.*" />
  <PackageReference Include="NLog" Version="4.*" />
</ItemGroup>
```

snakefoot (approved):
Looks great, only have some minor suggestions. And thank you so much for the Wiki-page. Excellent job.

examples/ASP.NET Core 6/ASP.NET Core 6 NLog Example/nlog.config
ThomasArdal: Sure thing. The `nlog.config` file in the ASP.NET Core 6 sample is a copy from the ASP.NET Core 5 sample with only the paths updated. I can upgrade it in the sample if you prefer the layout from the Wiki. We should probably do that for the ASP.NET Core 5 sample too, right?

snakefoot: Yes and that change is already pending with #734

appveyor.yml
ThomasArdal: I had to for the code to compile against .NET 6 👍

examples/ASP.NET Core 6/ASP.NET Core 6 NLog Example/Program.cs
snakefoot: Yes leave the docs with `4.*` until we have NLog 5.0 RTM on the road.

snakefoot:
@ThomasArdal I approve the wiki page. It is a nice christmas present.

ThomasArdal:
🎅

ThomasArdal:
I believe all of the wiki pages are updated now.

304NotModified:
Close/reopen to trigger CI (Azure Devops)

304NotModified:
Thanks a lot!

src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
```diff
@@ -72,6 +73,10 @@ NLog 5 release post: https://nlog-project.org/2021/08/25/nlog-5-0-preview1-ready
     <Title>ASP.NET Core 5 integration for NLog - .NET 5.0</Title>
     <DefineConstants>$(DefineConstants);ASP_NET_CORE;ASP_NET_CORE3</DefineConstants>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'net6.0' ">
```
304NotModified, line 76: FYI I mark this as feature

Issue #753, opened 2022-02-01
Multiple ASP.NET core renderers bug


tometchy:
**Type**: Bug

**NLog version**: 4.7.13

**NLog.Web.AspNetCore version**: 4.14.0 

**NLog.Extensions.Logging version**: --

**Platform**: .Net 6

**Current NLog config**:
``` xml
<?xml version="1.0" encoding="utf-8" ?>
<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      autoReload="true"
      internalLogLevel="Info"
      internalLogFile="c:\temp\internal-nlog-AspNetCore.txt">

  <!-- enable asp.net core layout renderers -->
  <extensions>
    <add assembly="NLog.Web.AspNetCore"/>
  </extensions>

  <!-- the targets to write to -->
  <targets>
    <target xsi:type="Console" name="console" layout="${callsite}${aspnet-request-url}${aspnet-request-posted-body}" />
  </targets>

  <rules>
    <logger name="*" minlevel="Trace" writeTo="console" />
  </rules>
</nlog>
```

Csproj:
```xml
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <Nullable>enable</Nullable>
    <ImplicitUsings>enable</ImplicitUsings>
  </PropertyGroup>
    
  <ItemGroup>
    <PackageReference Include="NLog" Version="4.7.13" />
    <PackageReference Include="NLog.Web.AspNetCore" Version="4.14.0" />
  </ItemGroup>

</Project>
```

Program.cs:
``` csharp
using NLog;
using NLog.Web;
using Microsoft.AspNetCore.Mvc;

var logger = NLog.LogManager.Setup().LoadConfigurationFromAppSettings().GetCurrentClassLogger();
logger.Debug("init main");

try
{
    var builder = WebApplication.CreateBuilder(args);
    
    builder.Services.AddControllersWithViews();

    builder.Logging.ClearProviders();
    builder.Logging.SetMinimumLevel(Microsoft.Extensions.Logging.LogLevel.Trace);
    builder.Host.UseNLog();

    var app = builder.Build();

    if (!app.Environment.IsDevelopment())
    {
        app.UseExceptionHandler("/Home/Error");
[...]

PR #754, opened 2022-02-03
Removed ${aspnet-request-posted-body} because not threadsafe

snakefoot:
NLog 5.0 only supports threadsafe layoutrenderers. This LayoutRenderer modifies the Body-Stream while reading, so it gives unwanted side-effects and unexpected errors. See also #753

Instead request-body should be explicitly captured using the middleware-pipeline. And maybe stored in HttpContext.Items or in ScopeContext-Property.


snakefoot:
Thanks to @bakgerman now NLog.Web.AspNetCore v5.1 re-introduces [${aspnet-request-posted-body}](https://github.com/NLog/NLog/wiki/AspNet-Request-posted-body-layout-renderer) with help from middleware:
```c#
app.UseMiddleware<NLog.Web.NLogRequestPostedBodyMiddleware>();
```
It is no longer necessary to call `context.Request.EnableBuffering();` as it is handled by the middleware.

PR #781, opened 2022-05-24
Hopefully restore aspnet-request-posted-body using middleware approac…

bakgerman:
…h and HttpContext.Items collection

### DIFF
--- a/src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
+++ b/src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
@@ -1,0 +1,138 @@
+﻿using System.IO;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using NLog.Common;
+using NLog.Web.LayoutRenderers;
+
+namespace NLog.Web
+{
+    /// <summary>
+    /// This class is to intercept the HTTP pipeline and to allow additional logging of the following
+    ///
+    /// POST request body
+    ///
+    /// The following are saved in the HttpContext.Items collection
+    ///
+    /// __nlog-aspnet-request-posted-body
+    ///
+    /// Usage: app.UseMiddleware&lt;NLogRequestPostBodyMiddleware&gt;(); where app is an IApplicationBuilder
+    /// Register the NLogRequestPostBodyMiddlewareConfiguration in the IoC so that the config gets passed to the constructor
+    /// </summary>
+    public class NLogRequestPostedBodyMiddleware : IMiddleware
+    {
+        private NLogRequestPostedBodyMiddlewareConfiguration _configuration { get; }
+
+        /// <summary>
+        /// Constructor that takes a configuration
+        /// </summary>
+        /// <param name="configuration"></param>
+        public NLogRequestPostedBodyMiddleware(NLogRequestPostedBodyMiddlewareConfiguration configuration)
+        {
+            _configuration = configuration;
+        }
+
+        /// <summary>
+        /// This allows interception of the HTTP pipeline for logging purposes
+        /// </summary>
+        /// <param name="context">The HttpContext</param>
+        /// <param name="next">The RequestDelegate that is to be executed next in the HTTP pipeline</param>
+        /// <returns></returns>
+        public async Task InvokeAsync(HttpContext context, RequestDelegate next)
+        {
+            // Perform null checking
+            if (context.Request == null)
+            {
+                InternalLogger.Debug("NLogRequestPostedBodyMiddleware: HttpContext.Request stream is null");
+                // Execute the next class in the HTTP pipeline, this can be the next middleware or the actual handler
+                await next(context).ConfigureAwait(false);
+                return;
+            }
+
+            // Perform null checking
+            if (context.Request.Body == null)
+            {
+                InternalLogger.Debug("NLogRequestPostedBodyMiddleware: HttpContext.Request.Body stream is null");
+                // Execute the next class in the HTTP pipeline, this can be the next middleware or the actual handler
+                await next(context).ConfigureAwait(false);
+                return;
+            }
+
+            // If we cannot read the stream we cannot capture the body
+            if (!context.Request.Body.CanRead)
+            {
+                InternalLogger.Debug("NLogRequestPostedBodyMiddleware: HttpContext.Request.Body stream is non-readable");
+                // Execute the next class in the HTTP pipeline, this can be the next middleware or the actual handler
+                await next(context).ConfigureAwait(false);
+                return;
+            }
+
+            // If we cannot reset the stream position to zero, and then back to the original position
+            // we cannot capture the body
+            if (!context.Request.Body.CanSeek)
+            {
+                InternalLogger.Debug("NLogRequestPostedBodyMiddleware: HttpContext.Request.Body stream is non-seekable");
+                // Execute the next class in the HTTP pipeline, this can be the next middleware or the actual handler
+                await next(context).ConfigureAwait(false);
+                return;
+            }
+
+            // This is required, otherwise reading the request will destructively read the request
+            context.Request.EnableBuffering();
+
+            // Use the predicate in the configuration instance that takes the HttpContext as an argument
+            if (_configuration.ShouldCapture(context))
+            {
+                // Save the POST request body in HttpContext.Items with a key of '__nlog-aspnet-request-posted-body'
+                context.Items[AspNetRequestPostedBodyLayoutRenderer.NLogPostedRequestBodyKey] =
+                    await GetString(context?.Request.Body).ConfigureAwait(false);
+            }
+            else
+            {
+                InternalLogger.Debug("NLogRequestPostedBodyMiddleware: _configuration.ShouldCapture(HttpContext) predicate returned false");
+            }
+
+            // Execute the next class in the HTTP pipeline, this can be the next middleware or the actual handler
+            await next(context).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Convert the stream to a String for logging.
+        /// If the stream is binary please do not utilize this middleware
+        /// Arguably, logging a byte array in a sensible format is simply not possible.
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <returns>The contents of the Stream read fully from start to end as a String</returns>
+        private async Task<string> GetString(Stream stream)
+        {
+            // Save away the original stream position
+            var originalPosition = stream.Position;
+
+            // This is required to reset the stream position to the beginning in order to properly read all of the stream.
+            stream.Position = 0;
+
+            string responseText = null;
+
+            // The last argument, leaveOpen, is set to true, so that the stream is not pre-maturely closed
+            // therefore preventing the next reader from reading the stream.
+            // The middle three arguments are from the configuration instance
+            // These default to UTF-8, true, and 1024.
+            using (var streamReader = new StreamReader(
+                       stream,
+                       _configuration.Encoding,
+                       _configuration.DetectEncodingFromByteOrderMark,
+                       _configuration.BufferSize,
+                       leaveOpen: true))
+            {
+                // This is the most straight forward logic to read the entire body
+                responseText = await streamReader.ReadToEndAsync().ConfigureAwait(false);
+            }
+
+            // This is required to reset the stream position to the original, in order to
+            // properly let the next reader process the stream from the original point
+            stream.Position = originalPosition;
+
+            // Return the string of the body
+            return responseText;
+        }
+    }
+}
--- a/src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddlewareConfiguration.cs
+++ b/src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddlewareConfiguration.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Text;
+using Microsoft.AspNetCore.Http;
+
+namespace NLog.Web
+{
+    /// <summary>
+    /// Contains the configuration for the NLogRequestPostedBodyMiddleware
+    /// </summary>
+    public class NLogRequestPostedBodyMiddlewareConfiguration
+    {
+        /// <summary>
+        /// The default configuration
+        /// </summary>
+        public static readonly NLogRequestPostedBodyMiddlewareConfiguration Default = new NLogRequestPostedBodyMiddlewareConfiguration();
+
+        /// <summary>
+        /// Defaults to UTF-8
+        /// </summary>
+        public Encoding Encoding { get; set; } = Encoding.UTF8;
+
+        /// <summary>
+        /// Defaults to 1024
+        /// </summary>
+        public int BufferSize { get; set; } = 1024;
+
+        /// <summary>
+        /// Defaults to true
+        /// </summary>
+        public bool DetectEncodingFromByteOrderMark { get; set; } = true;
+
+        /// <summary>
+        /// If this returns true, the post request body will be captured
+        /// Defaults to true if content length &lt;= 8192
+        /// This can be used to capture only certain content types,
+        /// only certain hosts, only below a certain request body size, and so forth
+        /// </summary>
+        /// <returns></returns>
+        public Predicate<HttpContext> ShouldCapture { get; set; } = DefaultCapture;
+
+        /// <summary>
+        /// The default predicate for ShouldCapture
+        /// Returns true if content length &lt;= 8192
+        /// </summary>
+        public static bool DefaultCapture(HttpContext context)
+        {
+            return context?.Request?.ContentLength != null && context?.Request?.ContentLength <= 8192;
+        }
+    }
+}
--- a/src/NLog.Web/NLogRequestPostedBodyHttpModule.cs
+++ b/src/NLog.Web/NLogRequestPostedBodyHttpModule.cs
@@ -1,0 +1,135 @@
+﻿using System;
+using System.Collections;
+using System.IO;
+#if NET46_OR_GREATER
+using System.Text;
+#endif
+using System.Web;
+using NLog.Web.LayoutRenderers;
+
+namespace NLog.Web
+{
+    /// <summary>
+    /// This class is to intercept the HTTP pipeline and to allow additional logging of the following
+    ///
+    /// POST request body
+    ///
+    /// The following are saved in the HttpContext.Items collection
+    ///
+    /// __nlog-aspnet-request-posted-body
+    /// </summary>
+    public class NLogRequestPostedBodyHttpModule : IHttpModule
+    {
+        /// <summary>
+        /// The name of the HttpModule
+        /// </summary>
+        public string ModuleName => "NLog Request Posted Body Module";
+
+        /// <summary>
+        /// Hook in to the BeginRequest event to capture the request posted body
+        /// </summary>
+        /// <param name="context"></param>
+        public void Init(HttpApplication context)
+        {
+            context.BeginRequest += InterceptRequest;
+        }
+
+        /// <summary>
+        /// This will forward the necessary arguments to the capture request body method
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="args"></param>
+        protected void InterceptRequest(object sender, EventArgs args)
+        {
+            HttpApplication app = sender as HttpApplication;
+
+            CaptureRequestPostedBody(app?.Request?.InputStream, app?.Request?.ContentLength, app?.Context?.Items);
+        }
+
+        /// <summary>
+        /// This will capture the requested posted body if &lt;= 8192 bytes
+        /// Public to be unit testable, HttpContext and HttpRequest are un-mockable
+        /// unless you are using ASP.NET Core.  HttpContext and HttpRequest are sealed
+        /// and no not have an interface so NSubstitute throws an Exception mocking them.
+        /// </summary>
+        /// <param name="bodyStream"></param>
+        /// <param name="contentLength"></param>
+        /// <param name="items"></param>
+        public void CaptureRequestPostedBody(Stream bodyStream, int? contentLength, IDictionary items)
+        {
+            if (bodyStream == null)
+            {
+                return;
+            }
+
+            if (!bodyStream.CanRead)
+            {
+                return;
+            }
+
+            if (!bodyStream.CanSeek)
+            {
+                return;
+            }
+
+            if (contentLength == null || contentLength > 8192)
+            {
+                return;
+            }
+
+            items[AspNetRequestPostedBodyLayoutRenderer.NLogPostedRequestBodyKey] = GetString(bodyStream);
+        }
+
+        /// <summary>
+        /// Reads the posted body stream into a string
+        /// </summary>
+        /// <param name="stream"></param>
+        /// <returns></returns>
+        protected string GetString(Stream stream)
+        {
+            string responseText = string.Empty;
+
+            // Save away the original stream position
+            var originalPosition = stream.Position;
+
+            // This is required to reset the stream position to the beginning in order to properly read all of the stream.
+            stream.Position = 0;
+
+#if NET46_OR_GREATER
+
+            //This required 5 argument constructor with leaveOpen available was added in 4.5,
+            //but the project and its unit test project are built for 4.6
+            //and we should not change the csproj file just for this single class
+
+            //If the 4 argument constructor with leaveOpen missing is used, the stream is closed after the
+            //ReadToEnd() operation completes and the request stream is no longer open for the actual consumer
+            //This causes the unit test to fail and should cause a failure during actual usage.
+
+            using (var streamReader = new StreamReader(
+                       stream,
+                       Encoding.UTF8,
+                       detectEncodingFromByteOrderMarks: true,
+                       bufferSize: 1024,
+                       leaveOpen: true))
+            {
+                // This is the most straight forward logic to read the entire body
+                responseText = streamReader.ReadToEnd();
+            }
+#endif
+            // This is required to reset the stream position to the original, in order to
+            // properly let the next reader process the stream from the original point
+            stream.Position = originalPosition;
+
+            // Return the string of the body
+            return responseText;
+        }
+
+        /// <summary>
+        /// This is a no-op
+        /// </summary>
+        public void Dispose()
+        {
+            // Nothing to be disposed of
+        }
+    }
+}
--- a/src/Shared/LayoutRenderers/AspNetRequestPostedBodyLayoutRenderer.cs
+++ b/src/Shared/LayoutRenderers/AspNetRequestPostedBodyLayoutRenderer.cs
@@ -1,0 +1,70 @@
+﻿using System.Text;
+using NLog.LayoutRenderers;
+#if ASP_NET_CORE
+using Microsoft.AspNetCore.Http;
+#else
+using System.Web;
+#endif
+namespace NLog.Web.LayoutRenderers
+{
+    /// <summary>
+    /// ASP.NET posted body, e.g. FORM or Ajax POST
+    /// This this only works on .NET 4.6 or above due to a required
+    /// StreamReader constructor not being available
+    /// </summary>
+    /// <para>Example usage of ${aspnet-request-posted-body}:</para>
+    /// <example>
+    /// <code lang="NLog Layout Renderer">
+    /// ${aspnet-request-posted-body} - Produces - {username:xyz,password:xyz}
+    /// </code>
+    /// </example>
+    [LayoutRenderer("aspnet-request-posted-body")]
+    public class AspNetRequestPostedBodyLayoutRenderer : AspNetLayoutRendererBase
+    {
+
+        /// <summary>
+        /// The string for the key in HttpContext.Items for the POST request body
+        /// </summary>
+        public static readonly string NLogPostedRequestBodyKey = "__nlog-aspnet-request-posted-body";
+
+        /// <summary>Renders the ASP.NET posted body</summary>
+        /// <param name="builder">The <see cref="T:System.Text.StringBuilder" /> to append the rendered data to.</param>
+        /// <param name="logEvent">Logging event.</param>
+        protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
+        {
+            var httpContext = HttpContextAccessor.HttpContext;
+            if (httpContext == null)
+            {
+                return;
+            }
+
+            var items = httpContext.Items;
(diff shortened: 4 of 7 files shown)


bakgerman:
Attempt to restore aspnet-request-posted-body layout renderer in NLog.Web 5.0.

snakefoot:
Thank you for this great contribution. Really amazing that you include both NLog.Web (HttpModule) and NLog.Web.AspNetCore (Middleware). I like the idea of using delegates, that allows the users to plugin functionality without having to inherit and override methods.

Would it make sense if we created one AspNetCore-middleware for NLog, where one of the features is correctly capturing `${aspnet-request-posted-body}` ? Was thinking to merge it together with NLogRequestLoggingMiddleware, so request-body capture is one extra feature (And change NLogRequestLoggingMiddleware so its current exception-/error-logging also become plugin-style).

Was also thinking to make the Enable-Buffering-Condition-Logic a little "smarter", since the `EnableBuffering()`-buffering is super expensive. So maybe check the ContentType / ContentLength before actually doing the buffering. And if the buffering is below 64 KByte then just use a byte-array to avoid using the heavy tmp-file-buffering implemented by `EnableBuffering()`

bakgerman:
Thank you, I worked very hard on the PR.  :)

If you want a single middleware class that is clearly the proper goal, I did not want to change an existing class since this is my first commit to NLog.Web, but yes i thought when I was doing the PR that having > 1 middleware is not the best idea.
Can you please explain 'become plugin style' in some more detail before I feel I can make the change?

I can do the EnableBuffering() only if < 64K change quickly, later today or tomorrow. 
I will have to test it thru an actual http flow as I recall sometimes bad things happened if this was off, but that ight have been only for > 64K case.

I have used NLog heavily and would like to help enhance one of my favorite github projects.

snakefoot:
> If you want a single middleware class that is clearly the proper goal, I did not want to change an existing class since this is my first commit to NLog.Web

I was also guessing that was your approach (to keep things simple and isolated before interfering with existing classes).

> Can you please explain 'become plugin style' in some more detail before I feel I can make the change?

You have created `NLogRequestPostedBodyMiddlewareConfiguration` where one can plugin custom delegates, that will be used instead of the default-delegate.

I was thinking a similar thing could be done for `NLogRequestLoggingMiddleware`, so it calls a plugin-method defined on `NLogRequestLoggingOptions` that decides whether to do exception-/error-logging.

bakgerman:
```
"You have created NLogRequestPostedBodyMiddlewareConfiguration where one can plugin custom delegates, that will be used instead of the default-delegate.

I was thinking a similar thing could be done for NLogRequestLoggingMiddleware, so it calls a plugin-method defined on NLogRequestLoggingOptions that decides how it should perform exception-/error-logging (when assigned null then it is skipped)."
```

I see what you mean now.

For the EnableBuffering(), it seems like Microsoft has already done the in memory part for us:
"Normally buffers request bodies in memory; writes requests larger than 30K bytes to disk."

As I am using my middleware for rest and soap, i do not have a test case > 30K.
We did test the performance of the middleware and async NLog.Database logging, with the commercial DynaTrace tool, and it was found to not harm any performance.

snakefoot:
> For the EnableBuffering(), it seems like Microsoft has already done the in memory part for us:
> > "Normally buffers request bodies in memory; writes requests larger than 30K bytes to disk."

Then I think 30K should be the default-max-limit for NLog-midddleware, and if ContentLength is larger, then it will not attempt `EnableBuffering()`.

bakgerman:
Well, I found that if I leave EnableBuffing() off, stream.CanSeek returns false and we cannot capture the request body since we have to reset Position property to zero, read to end, then rest the Position back to original.  The unit test does not show this but actual end to end test does show this.

bakgerman:
ok I will make limit = 30k instead of 8 k

bakgerman:
Default limit changed from 8KB to 30KB per Microsoft comments on EnableBuffering() method.

bakgerman:
Will need help testing the IHttpModule version end to end, I do not have any legacy 4.6.2 Framework web application anymore at my team where I work.

bakgerman:
Will examine squeezing 2 middleware into 1 class later this week.

bakgerman:
I think I am ready for the next review.  The last commit unit tests failed due to an unrelated nuget.org error and I am researching how to rerun the checks for the last commit.

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
```diff
@@ -0,0 +1,138 @@
...
+            if (!context.Request.Body.CanRead)
+            {
+                InternalLogger.Debug("NLogRequestPostedBodyMiddleware: HttpContext.Request.Body stream is non-readable");
+                // Execute the next class in the HTTP pipeline, this can be the next middleware or the actual handler
+                await next(context).ConfigureAwait(false);
+                return;
+            }
+
+            // This is required, otherwise reading the request will destructively read the request
+            context.Request.EnableBuffering();
```
snakefoot, line 70: I think `EnableBuffering` should be moved inside `if (_configuration.ShouldCapture(context))`

src/NLog.Web/NLogRequestPostedBodyHttpModule.cs
```diff
@@ -0,0 +1,169 @@
...
+
+            if (!bodyStream.CanSeek)
+            {
+                InternalLogger.Debug("NLogRequestPostedBodyMiddleware: HttpContext.Request.InputStream stream is non-seekable");
+                return;
+            }
+
+            if (shouldCapture)
+            {
+                items[AspNetRequestPostedBodyLayoutRenderer.NLogPostedRequestBodyKey] = GetString(bodyStream);
```
snakefoot, line 94: When `GetString(bodyStream)` returns `null`  or `string.Empty`, then it should not insert into items

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// <summary>
+        /// This allows interception of the HTTP pipeline for logging purposes
+        /// </summary>
+        /// <param name="context">The HttpContext</param>
+        /// <param name="next">The RequestDelegate that is to be executed next in the HTTP pipeline</param>
+        /// <returns></returns>
+        public async Task InvokeAsync(HttpContext context, RequestDelegate next)
+        {
+            // Perform null checking
+            if (context.Request == null)
```
snakefoot, line 43: Could all these `if`-statements be moved into a private-methods, so it becomes like this:

```c#
if (!ShouldCaptureRequestBody(context))
{
    await next(context).ConfigureAwait(false);
    return;
}
```

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
```diff
@@ -0,0 +1,138 @@
...
+                await next(context).ConfigureAwait(false);
+                return;
+            }
+
+            // Use the predicate in the configuration instance that takes the HttpContext as an argument
+            if (_configuration.ShouldCapture(context))
+            {
+                // Save the POST request body in HttpContext.Items with a key of '__nlog-aspnet-request-posted-body'
+                context.Items[AspNetRequestPostedBodyLayoutRenderer.NLogPostedRequestBodyKey] =
+                    await GetString(context?.Request.Body).ConfigureAwait(false);
```
snakefoot, line 87: When `GetString(bodyStream)` returns null or string.Empty, then it should not insert into items

src/NLog.Web/NLogRequestPostedBodyHttpModule.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            context.BeginRequest += InterceptRequest;
+        }
+
+        /// <summary>
+        /// This will forward the necessary arguments to the capture request body method
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="args"></param>
+        protected void InterceptRequest(object sender, EventArgs args)
```
snakefoot, line 51: Could this method become private and renamed to `BeginRequest`. I prefer to keep the visibility of methods to absolute minimum, until someone needs access.

src/NLog.Web/NLogRequestPostedBodyHttpModule.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            HttpApplication app = sender as HttpApplication;
+
+            CaptureRequestPostedBody(
```
snakefoot, line 55: Could this become

```c#
if (ShouldCaptureRequestPostedBody(app))
{
      CaptureRequestPostedBody(app);
}
```

src/NLog.Web/NLogRequestPostedBodyMiddlewareConfiguration.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <returns></returns>
+        public Predicate<HttpApplication> ShouldCapture { get; set; } = DefaultCapture;
+
+        /// <summary>
+        /// The default predicate for ShouldCapture
+        /// Returns true if content length &lt;= 30KB
+        /// </summary>
+        public static bool DefaultCapture(HttpApplication app)
+        {
+            return app?.Context?.Request?.ContentLength != null && app?.Context?.Request?.ContentLength <= 30 * 1024;
```
snakefoot, line 47: Could `30 * 1024` become a setting ?

src/NLog.Web/NLogRequestPostedBodyMiddlewareConfiguration.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// Defaults to UTF-8
+        /// </summary>
+        public Encoding Encoding { get; set; } = Encoding.UTF8;
+
+        /// <summary>
+        /// Defaults to 1024
+        /// </summary>
+        public int BufferSize { get; set; } = 1024;
```
snakefoot, line 25: Not sure this should be a setting, until someone requests this. I prefer to keep the visibility of methods and properties to absolute minimum, until someone needs access.

src/NLog.Web/NLogRequestPostedBodyMiddlewareConfiguration.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        /// <summary>
+        /// The default configuration
+        /// </summary>
+        public static readonly NLogRequestPostedBodyMiddlewareConfiguration Default = new NLogRequestPostedBodyMiddlewareConfiguration();
+
+        /// <summary>
+        /// Defaults to UTF-8
+        /// </summary>
+        public Encoding Encoding { get; set; } = Encoding.UTF8;
```
snakefoot, line 20: Not sure this should be a public-setting, until someone requests this. I prefer to keep the visibility of methods and properties to absolute minimum, until someone needs access.

bakgerman: Removed, set to UTF8 in the NLogRequestPostedBpdyMiddleware class

src/NLog.Web/NLogRequestPostedBodyMiddlewareConfiguration.cs
bakgerman: Removed set to 1024 in the NLogRequestPostedBodyMiddleware class

src/NLog.Web/NLogRequestPostedBodyMiddlewareConfiguration.cs
bakgerman: Made a public property named MaximumRequestSize defaulting to 30KB.

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
bakgerman: Completed as requested

src/NLog.Web/NLogRequestPostedBodyHttpModule.cs
bakgerman: Completed as requested

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
bakgerman: Completed as requested

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
bakgerman: completed as requested

src/NLog.Web/NLogRequestPostedBodyHttpModule.cs
bakgerman: Completed as requested

src/NLog.Web/NLogRequestPostedBodyHttpModule.cs
bakgerman: I am working on this today.

bakgerman: I finished this, but encountered problems when unit testing.  I then make the private methods internal, and try to use InternalsVisibleTo attribute, but was unable to find the proper value for the Public Key.  The HttpContext and HttpRequest are not mockable in .NET 3.x and 4.x as they are sealed, and NSubstitute throws an exception attempting to mock them, that is why i tried to make the private methods internal so I can pass them arguments like MemoryStream directly from the unit tests.  The unit tests for this class are currently commented until this is resolved.

bakgerman:
AppVeyor seems to fail for random reasons un-related to my commits?

src/NLog.Web/NLogRequestPostedBodyHttpModule.cs
snakefoot: If it hurts unit-test coverage, then just keep them public, then I can try and figure out why InternalsVisibleTo is not working.

snakefoot:
Could have been nice to have HttpModule included as well, but understand how no unit-coverage and no testing-environment makes it less ideal.

Think NLogRequestPostedBodyMiddleware looks nice and simple. Would it make sense to merge it together with NLogRequestLoggingMiddleware?

I can refactor NLogRequestLoggingMiddleware, so it become optional whether to perform request-logging.

snakefoot:
Btw. any reason why you use `IMiddleware`, since it sounds like it has an [overhead](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/extensibility):

- Activation per client request (injection of scoped services)
- Strong typing of middleware

The `Activation` sounds like it will create an instance of the middleware for every client-request to allow constructor-injection of "scoped services")

bakgerman:
> Btw. any reason why you use `IMiddleware`, since it sounds like it has an [overhead](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/extensibility):
> 
> * Activation per client request (injection of scoped services)
> * Strong typing of middleware
> 
> The `Activation` sounds like it will create an instance of the middleware for every client-request to allow possible constructor-parameter-injection of "scoped services")

No, I would not like it to be constructed each time.  For now I have changed to have same structure as NLogRequestLoggingMiddleware.  I will have to fix that in the copy in my code at employer as well.  
I have changed the unit tests and I will commit that change.

bakgerman:
> Thank you for resolving all my comments.
> 
> Could have been nice to have HttpModule included as well, but understand how no unit-coverage and no testing-environment makes it less ideal.
> 
We can still include HttpModule version, but we must ask users for help in testing that it works and maybe should mark it experimental.  Let me know if you want it restored.

I still have additional layout renderers to PR, i went thru both Framework and Core HttpContexts to see what is not in NLog.Web, but are common in both Core and non-Core:

Request Client Certificate

Response Content Type
Response Cookies - requires Microsoft.AspNetCore.Http.Extensions NuGet in Asp.Net Core version, otherwise you can only add or delete cookie, but not read them
Response Headers

Remote Port
Local Address
Local Port

IsWebSocketRequest
WebSocketNegotiatedProtocol
WebSocketRequestedProtocols

ResponseBodyMiddlware and Layout Renderer - This one works at my employment fine, but we have <8K responses
This one, because it must write to MemoryStream, then be copied to HttpContext.Items, then memoryStream.CopyTo(actual stream back to browser or caller) always use 2x the memory.  So use carefully or we can not merge in at all.  Due to the fact that you must decide to capture BEFORE you have process the response, you cannot decide to skip response > 30K like in request body middleware where request exist BEFORE you decide to capture.

But the rest of the layout renders above are simple and just work like the rest already there.

src/Shared/LayoutRenderers/AspNetRequestPostedBodyLayoutRenderer.cs
```diff
@@ -0,0 +1,68 @@
...
+                return;
+            }
+#else
+            if (!items.ContainsKey(NLogPostedRequestBodyKey))
+            {
+                return;
+            }
+#endif
+
+            if (items[NLogPostedRequestBodyKey] is string)
```
snakefoot, line 62: Could this be changed to:
```c#
var postedBody = items[NLogPostedRequestBodyKey] as string;
if (string.IsNullOrEmpty(postedBody))
{
   builder.Append(postedBody);
}
```
To skip a dictionary-lookup and cast-check

snakefoot:
After pondering some more, then I think it is better not to merge `NLogRequestLoggingMiddleware` and `NLogRequestPostedBodyMiddleware`.

If necessary then one can make an extension method that includes both in one operation. Ex. want request-logging together with posted-body.

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
```diff
@@ -0,0 +1,133 @@
...
+    /// __nlog-aspnet-request-posted-body
+    ///
+    /// Usage: app.UseMiddleware&lt;NLogRequestPostBodyMiddleware&gt;(); where app is an IApplicationBuilder
+    /// Register the NLogRequestPostBodyMiddlewareConfiguration in the IoC so that the config gets passed to the constructor
+    /// </summary>
+    public class NLogRequestPostedBodyMiddleware
+    {
+        private readonly RequestDelegate _next;
+
+        private NLogRequestPostedBodyMiddlewareConfiguration _configuration { get; }
```
snakefoot, line 26: Can this become a readonly-field instead of a property ?

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <returns></returns>
+        public async Task Invoke(HttpContext context)
+        {
+            if (ShouldCaptureRequestBody(context))
+            {
+                // This is required, otherwise reading the request will destructively read the request
+                context.Request.EnableBuffering();
+
+                // Save the POST request body in HttpContext.Items with a key of '__nlog-aspnet-request-posted-body'
+                var requestBody = await GetString(context?.Request.Body).ConfigureAwait(false);
```
snakefoot, line 52: Can you remove the "?" so it just becomes `context.Request.Body` ?

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+            // Execute the next class in the HTTP pipeline, this can be the next middleware or the actual handler
+            await _next(context).ConfigureAwait(false);
+        }
+
+        private bool ShouldCaptureRequestBody(HttpContext context)
+        {
+            // Perform null checking
+            if (context.Request == null)
```
snakefoot, line 67: Can this become `context?.Request == null` ? (With "?")

src/Shared/LayoutRenderers/AspNetRequestPostedBodyLayoutRenderer.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// </code>
+    /// </example>
+    [LayoutRenderer("aspnet-request-posted-body")]
+    public class AspNetRequestPostedBodyLayoutRenderer : AspNetLayoutRendererBase
+    {
+
+        /// <summary>
+        /// The string for the key in HttpContext.Items for the POST request body
+        /// </summary>
+        public static readonly string NLogPostedRequestBodyKey = "__nlog-aspnet-request-posted-body";
```
snakefoot, line 26: Could this become `internal` instead of `public` ?

snakefoot: Any reason why you use a string-key with slow hashcode, instead of standard object:
```c#
internal static readonly object NLogPostedRequestBodyKey = new object();
```

src/Shared/LayoutRenderers/AspNetRequestPostedBodyLayoutRenderer.cs
bakgerman: Complete

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
bakgerman: Complete

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
bakgerman: Complete

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
bakgerman: Complete with ?.

src/Shared/LayoutRenderers/AspNetRequestPostedBodyLayoutRenderer.cs
bakgerman: Made internal.

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
bakgerman: maybe is better with 2 separate if statement so that InternalLogger.Debug has 2 separate messages

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
bakgerman: yes the other layout renders will be several small PRs after this is complete

src/Shared/LayoutRenderers/AspNetRequestPostedBodyLayoutRenderer.cs
bakgerman: Converted to object as requested.

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddlewareConfiguration.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        /// <summary>
+        /// The default configuration
+        /// </summary>
+        public static readonly NLogRequestPostedBodyMiddlewareConfiguration Default = new NLogRequestPostedBodyMiddlewareConfiguration();
+
+        /// <summary>
+        /// Defaults to true
+        /// </summary>
+        public bool DetectEncodingFromByteOrderMark { get; set; } = true;
```
snakefoot, line 19: Please remove this as setting, until a user needs to change it.

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddlewareConfiguration.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// only certain hosts, only below a certain request body size, and so forth
+        /// </summary>
+        /// <returns></returns>
+        public Predicate<HttpContext> ShouldCapture { get; set; } = DefaultCapture;
+
+        /// <summary>
+        /// The default predicate for ShouldCapture
+        /// Returns true if content length &lt;= 30KB
+        /// </summary>
+        public static bool DefaultCapture(HttpContext context)
```
snakefoot, line 40: Can this become a private-method (not static)? And instead of allocating a new `NLogRequestPostedBodyMiddlewareConfiguration` the active configuration?

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddlewareConfiguration.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <summary>
+        /// Defaults to true
+        /// </summary>
+        public bool DetectEncodingFromByteOrderMark { get; set; } = true;
+
+        /// <summary>
+        /// The maximum request size that will be captured
+        /// Defaults to 30KB
+        /// </summary>
+        public int MaximumRequestSize { get; set; } = 30 * 1024;
```
snakefoot, line 25: Can it be renamed to `MaxPostedBodySize` ? Maybe update comments about how it checks against `ContentLength` to protect against "very large" posted-body-payloads. And that EnableBuffering has an upper-limit of 30KByte where it changes to TEMP-files.

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddlewareConfiguration.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    /// <summary>
+    /// Contains the configuration for the NLogRequestPostedBodyMiddleware
+    /// </summary>
+    public class NLogRequestPostedBodyMiddlewareConfiguration
+    {
+        /// <summary>
+        /// The default configuration
+        /// </summary>
+        public static readonly NLogRequestPostedBodyMiddlewareConfiguration Default = new NLogRequestPostedBodyMiddlewareConfiguration();
```
snakefoot, line 14: Can this be changed from public to internal ?

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddlewareConfiguration.cs
```diff
@@ -0,0 +1,46 @@
+﻿using System;
+using Microsoft.AspNetCore.Http;
+
+namespace NLog.Web
+{
+    /// <summary>
+    /// Contains the configuration for the NLogRequestPostedBodyMiddleware
+    /// </summary>
+    public class NLogRequestPostedBodyMiddlewareConfiguration
```
snakefoot, line 9: Can it be renamed to NLogRequestPostedBodyMiddlewareOptions` ?

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddlewareConfiguration.cs
bakgerman: Complete

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddlewareConfiguration.cs
bakgerman: Complete

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddlewareConfiguration.cs
bakgerman: Complete

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
```diff
@@ -0,0 +1,136 @@
...
+        private readonly RequestDelegate _next;
+
+        private readonly NLogRequestPostedBodyMiddlewareOptions _configuration;
+
+        /// <summary>
+        /// Constructor that takes a configuration
+        /// </summary>
+        /// <param name="next"></param>
+        /// <param name="configuration"></param>
+        public NLogRequestPostedBodyMiddleware(RequestDelegate next, NLogRequestPostedBodyMiddlewareOptions configuration)
```
snakefoot, line 29: Can input-parameter `configuration` be renamed to `options` (together with `_configuration`)

snakefoot:
Thank you so much for being persistent and completing all my requests.

I will try to add the ASP.NET Classic support using HttpModule, at least as first attempt

bakgerman:
Thank you for all of the reviews, I know you are very busy with all of the NLog work!  If you need assistance with HttpModule, please let me know.  Hopefully the code in the middleware gives a direction to develop the HttpModule.  Microsoft makes it very hard to unit tests pre-ASP.NET Core middleware (meaning HttpModule) or HttpContext class.  Maybe I can find a legacy Framework ASP.NET app at my employer that I can test with, my direct department does not but the departments we work with may have one.

snakefoot:
@bakgerman Would it make sense to also have a Content-Type filter, so only trying to handle posted-body when in expected-formats (Avoid images / binaries / etc.)

```
application/json
application/xml
application/xhtml
text/javascript
text/html
text/plain
text/xml
```

snakefoot:
@bakgerman Created #787 with a very primitive ContentType-filter.

snakefoot:
Would it be possible for you to update this wiki-page, and describe the recommended way for enabling `${aspnet-request-posted-body}` using middleware:

https://github.com/NLog/NLog/wiki/AspNet-Request-posted-body-layout-renderer

You are very wellcome to remove all the old noise about previous version/attempts, and just state that one must use NLog.Web.AspNetCore v5.1 (or newer).

bakgerman:
Certainly

snakefoot:
Could you also add details about using the options for the middleware?


Sent from my Galaxy



Date: 18/07/2022 22:40 (GMT+01:00)
Cc: Rolf Kristensen ***@***.***>, State change ***@***.***>
Subject: Re: [NLog/NLog.Web] Restore aspnet-request-posted-body with middleware (PR #781)


I have changed it as requested.

bakgerman:
Sure.

Should we add the response body middleware as example to another page?

bakgerman:
Added a large section about the middleware options to the end of the page

snakefoot:
Looks great. Thank you again.

You are very welcome to add new wiki-page about logging the response body. Then add link from request-body-layout-wiki-page.


Sent from my Galaxy



Date: 18/07/2022 23:56 (GMT+01:00)
Cc: Rolf Kristensen ***@***.***>, State change ***@***.***>
Subject: Re: [NLog/NLog.Web] Restore aspnet-request-posted-body with middleware (PR #781)


Added a large section about the middleware options to the end of the page

PR #787, opened 2022-06-04
NLogRequestPostedBodyMiddleware - Only accept AllowContentTypes

snakefoot:


bakgerman:
Should we also add 'application/soap+xml' to allow content type list, or will the charset entry already allow this?  Many time the content type for SOAP is 'application/soap+xml; charset=utf-8'.

snakefoot:
> Should we also add 'application/soap+xml' to allow content type list, or will the charset entry already allow this? Many time the content type for SOAP is 'application/soap+xml; charset=utf-8'.

That should be covered by `application/`-prefix + `xml`-suffix. But yes any ContentType that contains `charset` will also pass the filter (expects charset means text that has been encoded).

PR #789, opened 2022-06-04
ASP.NET Response Cookie Layout Renderer

bakgerman:
Add Response Cookie Layout Render based on request Cookie layout render.
But, this has important difference for ASP.NET Core version.
We need to install Microsoft.AspNetCore.Http.Extensions NuGet package version 2.1.x other wise the cookie for response can only be Append() or Delete().
So, when we install this NuGet, it installs extension methods for Session named GetString() and GetInt32().
So these 2 method need to be delete from HttpContextExtensions.cs otherwise compiler gives 2 ambiguous call errors.
After these are deleted, re-executing all the unit test still give successful results.

### DIFF
--- a/src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
+++ b/src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
@@ -75,15 +75,14 @@ NLog 5 release post: https://nlog-project.org/2021/08/25/nlog-5-0-preview1-ready
     <DefineConstants>$(DefineConstants);ASP_NET_CORE;ASP_NET_CORE3</DefineConstants>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="2.1.21" />
-    <PackageReference Include="NLog.Extensions.Logging" Version="5.0.0" />
-    <PackageReference Include="System.Text.Encodings.Web" Version="4.5.1" />
+	  <PackageReference Include="NLog.Extensions.Logging" Version="5.0.0" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'net461' ">
     <!-- Fixed to 2.1.0 as 2.1 is Long Term Supported (LTS) and works with vanilla .NET Core 2.1 SDK -->
     <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.1.22" />
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.1.0" />
     <PackageReference Include="Microsoft.AspNetCore.Routing.Abstractions" Version="2.1.0" />
+    <PackageReference Include="System.Text.Encodings.Web" Version="4.5.1" />
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
--- a/src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
+++ b/src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
@@ -77,13 +77,13 @@ NLog 5 release post: https://nlog-project.org/2021/08/25/nlog-5-0-preview1-ready
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="2.1.21" />
     <PackageReference Include="NLog.Extensions.Logging" Version="5.0.0" />
+    <PackageReference Include="System.Text.Encodings.Web" Version="4.5.1" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'net461' ">
     <!-- Fixed to 2.1.0 as 2.1 is Long Term Supported (LTS) and works with vanilla .NET Core 2.1 SDK -->
     <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.1.22" />
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.1.0" />
     <PackageReference Include="Microsoft.AspNetCore.Routing.Abstractions" Version="2.1.0" />
-    <PackageReference Include="System.Text.Encodings.Web" Version="4.5.1" />
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="All" />
--- a/src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
+++ b/src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
@@ -75,11 +75,12 @@ NLog 5 release post: https://nlog-project.org/2021/08/25/nlog-5-0-preview1-ready
     <DefineConstants>$(DefineConstants);ASP_NET_CORE;ASP_NET_CORE3</DefineConstants>
   </PropertyGroup>
   <ItemGroup>
-	<PackageReference Include="NLog.Extensions.Logging" Version="5.0.0" />
+    <PackageReference Include="NLog.Extensions.Logging" Version="5.0.0" />
   </ItemGroup>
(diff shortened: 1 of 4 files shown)


src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
```diff
@@ -75,14 +75,15 @@ NLog 5 release post: https://nlog-project.org/2021/08/25/nlog-5-0-preview1-ready
     <DefineConstants>$(DefineConstants);ASP_NET_CORE;ASP_NET_CORE3</DefineConstants>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="2.1.21" />
```
snakefoot, line 78: Can we not extract the method needed for the Cookie-rendering, instead of adding as dependency ?

src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
```diff
@@ -75,14 +75,15 @@ NLog 5 release post: https://nlog-project.org/2021/08/25/nlog-5-0-preview1-ready
...
     <PackageReference Include="NLog.Extensions.Logging" Version="5.0.0" />
+    <PackageReference Include="System.Text.Encodings.Web" Version="4.5.1" />
```
snakefoot, line 80: Please move this nuget-dependency back to its original location.

src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
bakgerman: ok, let me work on that.

src/NLog.Web.AspNetCore/NLog.Web.AspNetCore.csproj
bakgerman: ok I will undo it

src/Shared/LayoutRenderers/AspNetResponseCookieLayoutRenderer.cs
```diff
@@ -0,0 +1,194 @@
...
+#elif ASP_NET_CORE2
+
+        /// <summary>
+        /// Method to wrap getting cookies from the HTTP Response for both Framework and Core
+        /// </summary>
+        /// <param name="response"></param>
+        /// <returns></returns>
+        protected IList<SetCookieHeaderValue> GetCookies(HttpResponse response)
+        {
+            var queryResults = response.Headers.Where(row => row.Key == "Set-Cookie").ToList();
```
snakefoot, line 147: Can one use `Microsoft.Net.Http.Headers.HeaderNames.SetCookie` ?

snakefoot: If `response.Headers` is type `IHeaderDictionary`. Can one not use `IHeaderDictionary.SetCookie`-property ?

snakefoot: So one just need to convert `SetCookie`-`StringValues` to something that `SetCookieHeaderValue.TryParse` can handle.

bakgerman: yes that is what I am working on now

bakgerman: I found IHeaderDictionary.SetCookie not exist in net std 2.0/net 461 case.

bakgerman: so i am working around that

bakgerman:
The 3 code smell come from the code I copy from RequestCookieLayoutRenderer.
I had to write GetCookies method only for ASP_NET_CORE2, since this is older version without SetCookie property on IHeaderDictionary  interface.  
The NuGet package changes are all rolled back to original.

snakefoot:
Will this not work for ASP_NET_CORE (net5 + netcoreapp3 + netstandard2), and will be faster than `GetTypedHeaders()` ?
```c#
        protected IList<SetCookieHeaderValue> GetCookies(HttpResponse response)
        {
            var queryResults = response.Headers[HeaderNames.SetCookie];
            if (queryResults.Count > 0 && SetCookieHeaderValue.TryParseList(queryResults, out var result))
                return result;
            else 
                return Array.Empty<SetCookieHeaderValue>();
        }
```

bakgerman:
Added back Microsoft.AspNetCore.http.Extensions NuGet Package, but ONLY for ASP_NET_CORE2 case.
This is a special case where the HttpContext.Response.Headers class is different from the ASP_NET_CORE3 case.
Copying the 1 line of the extension method required into our code failed since this uses a class that is not provided in ASP_NET_CORE2, so we need the whole NuGet package.  It is understood that the ASP_NET_CORE2 case has additional NuGets over ASP_NET_CORE3 case, so this should add the NuGet package ONLY where necessary and not in all cases.

bakgerman:
> Will this not work for ASP_NET_CORE (net5 + netcoreapp3 + netstandard2), and will be faster than `GetTypedHeaders()` ?
> 
> ```cs
>         protected IList<SetCookieHeaderValue> GetCookies(HttpResponse response)
>         {
>             var queryResults = response.Headers[HeaderNames.SetCookie];
>             if (queryResults.Count > 0 && SetCookieHeaderValue.TryParseList(queryResults, out var result))
>                 return result;
>             else 
>                 return Array.Empty<SetCookieHeaderValue>();
>         }
> ```

oh sorry I did not see, i will try that.  Did not realize there is parselist method.

src/Shared/LayoutRenderers/AspNetResponseCookieLayoutRenderer.cs
```diff
@@ -0,0 +1,180 @@
...
+        {
+            var httpResponse = HttpContextAccessor.HttpContext.TryGetResponse();
+            if (httpResponse == null)
+            {
+                return;
+            }
+
+            var cookies = GetCookies(httpResponse);
+#if ASP_NET_CORE
+            if (cookies.Any())
```
snakefoot, line 77: Can this be removed, so it just becomes `if (cookies.Count > 0)` for all platforms ?

bakgerman: Certainly

snakefoot: Now that the code-paths are identical. Can the `#if ASP_NET_CORE` be removed around `if (cookies.Count > 0)` ?

bakgerman: sorry about that, made the change too quickly

src/Shared/LayoutRenderers/AspNetResponseCookieLayoutRenderer.cs
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+#if !ASP_NET_CORE
+
+        /// <summary>
+        /// Method to get cookies for .NET Framework
+        /// </summary>
+        /// <param name="response"></param>
+        /// <returns></returns>
+        protected Cookies GetCookies(HttpResponseBase response)
```
snakefoot, line 89: Why protected instead of private ?

src/Shared/LayoutRenderers/AspNetResponseCookieLayoutRenderer.cs
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+
+#else
+        /// <summary>
+        /// Method to get cookies for all ASP.NET Core versions
+        /// </summary>
+        /// <param name="response"></param>
+        /// <returns></returns>
+        protected static IList<SetCookieHeaderValue> GetCookies(HttpResponse response)
```
snakefoot, line 147: Why protected ?

src/Shared/LayoutRenderers/AspNetResponseCookieLayoutRenderer.cs
bakgerman: Now private

src/Shared/LayoutRenderers/AspNetResponseCookieLayoutRenderer.cs
bakgerman: now private

snakefoot:
Thanks again for completing all my comments. And great work with the test-coverage.

snakefoot:
Created wiki page: https://github.com/NLog/NLog/wiki/AspNet-Response-Cookie-Layout-Renderer

PR #795, opened 2022-06-07
Response body capture middleware

bakgerman:


bakgerman:
This unit test that keeps failing needs to be fixed.  It makes 1 out of 5 of my commits break

src/Shared/Internal/HttpContextExtensions.cs
```diff
@@ -163,5 +163,31 @@ internal static bool HasAllowedContentType(this HttpContext context, IList<KeyVa
 
             return true;
         }
+
+        internal static bool HasAllowedResponseContentType(this HttpContext context, IList<KeyValuePair<string, string>> allowContentTypes)
```
snakefoot, line 167: Please improve the code-reuse here.

src/NLog.Web.AspNetCore/NLogRequestPostedBodyMiddleware.cs
```diff
@@ -59,8 +66,38 @@ public async Task Invoke(HttpContext context)
                 }
             }
 
-            // Execute the next class in the HTTP pipeline, this can be the next middleware or the actual handler
-            await _next(context).ConfigureAwait(false);
+            if (ShouldCaptureResponseBody(context))
```
snakefoot, line 69: Leaning toward having this feature in its own Middleware. Maybe in its own NLog-extension. Because of the `RecyclableMemoryStream`-dependency.

bakgerman: that is easy, it started in its own file, i can do that.

src/Shared/Internal/HttpContextExtensions.cs
bakgerman: of course

bakgerman:
well I think that just proves those unit tests for w3c render are not proper.

bakgerman:
Should this be a separate NuGet package due to RecycleableMemoryStream usage?

snakefoot:
> Should this be a separate NuGet package due to RecycleableMemoryStream usage?

Was actually thinking to make it a wiki-page-code-example-only, without any attempt to make code-reuse with `NLogRequestPostedBodyMiddleware`. 

Alternative it could become its own source-repository where it can grow and stabilize.

The reason is that I predict a lot of support questions around this feature, and not sure who will answer them.

bakgerman:
> 

I think wiki only is fine.  I am very agreeing with the lot of support questions comment.

snakefoot:
Thank you for being understanding. I prefer things that have no "hidden traps", and supporting unpredictable response-size can have side-effects

bakgerman:
I mean I think that is the only real solution available to us.  This feature is something that can be mis-used by the user without proper planning.

snakefoot:
When people see a hammer can solve a difficult problem, then they will get really angry when the hammer fails and it also hurt their foot.

Then they will no longer trust that hammer (for any problem), and will tell everyone they know not to use that hammer. Even if they used the hammer the wrong way, or used it for the wrong problem.

On the other hand, when people have built the hammer themselves, and it fails, then they will blame the problem for being too complex and hard.

bakgerman:
This is extremely very true!

bakgerman:
should we delete the PR and leave the branch so we can copy the code into wiki?  Do you want me to revert to standalone response body middleware?

snakefoot:
Think it would make sense to revert to the point just before the common `NLogMiddlewareOptions` is introduced.

And then just close the pull-request.

bakgerman:
Yes, that way the code will stand by itself and be ok for wiki

bakgerman:
OK, i think i have revert this back to the proper place with separate middleware and options class for response body

snakefoot:
Instead of doing revert commits, then you can just do a "hard reset" and then a "force push" for your pull-request-branch.

bakgerman:
I will do that, thank you

bakgerman:
Completed as Stated

snakefoot:
Guess it is ready to be closed?

bakgerman:
Yes I think so

PR #800, opened 2022-06-11
Added 7 layout renders for ITlsHandshakeFeature

bakgerman:
This is only for ASP_NET_CORE3.
This ITlsHandshakeFeature is not found in ASP_NET_CORE2.
(and obviously not in legacy 3.x or 4.x Framework...)
I did not yet write any unit tests.
If you think it is useful, please inform me to complete the unit tests and we can pull in 7 new renderers.

### DIFF
--- a/src/NLog.Web.AspNetCore/LayoutRenderers/AspNetTlsHandshakeCipherAlgorithm.cs
+++ b/src/NLog.Web.AspNetCore/LayoutRenderers/AspNetTlsHandshakeCipherAlgorithm.cs
@@ -1,0 +1,33 @@
+﻿using NLog.LayoutRenderers;
+using NLog.Web.Internal;
+using System.Text;
+
+namespace NLog.Web.LayoutRenderers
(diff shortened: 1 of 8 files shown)


snakefoot:
Think the idea is to only add the important stuff, where several people have a use-case.

Do not think the goal is to add as much as possible. Especially since any additional source-code has a cost in support-time (questions) and maintenance-time (api-behavior-changes). This weight that has to be carried, unless you want a library where half the functionality is not working, and causing users to be frustrated.

bakgerman:
I think that is correct, i only spend 45 minutes on it then i stopped and realized maybe nobody has asked for this.  But i got to learn more about .NET doing it.

Do you want some missing unit tests added instead?

snakefoot:
> Do you want some missing unit tests added instead?

Any improvements to test-code-coverage is always wellcome

bakgerman:
Closing this pull request

PR #805, opened 2022-06-23
Load config from json file with custom name

Fluttercom:
Extension method that loads configuration from json file with a custom name
See https://github.com/NLog/NLog/issues/4985

### DIFF
--- a/src/NLog.Web.AspNetCore/Config/SetupBuilderExtensions.cs
+++ b/src/NLog.Web.AspNetCore/Config/SetupBuilderExtensions.cs
@@ -20,7 +20,17 @@ namespace NLog.Web
         /// </summary>
         public static ISetupBuilder LoadConfigurationFromAppSettings(this ISetupBuilder setupBuilder, string basePath = null, string environment = null, string nlogConfigSection = "NLog", bool optional = true, bool reloadOnChange = false)
         {
+            return setupBuilder.LoadConfigurationFromJson("appsettings.json", basePath, environment, nlogConfigSection, optional, reloadOnChange);
+        }
+
+        /// <summary>
+        /// Loads NLog LoggingConfiguration from the NLog-section from json file with a custom name, respecting environment
+        /// </summary>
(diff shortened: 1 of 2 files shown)


snakefoot:
Think it should be a solution where `$[configsetting}` also works, with lookup in the `appsettings.json`

Fluttercom:
> 

Something like this?
`public static ISetupBuilder LoadConfigurationUsingConfigBuilder(this ISetupBuilder setupBuilder, Func<IConfigurationBuilder> builderFunc)`

snakefoot:
Yes together with string nlogConfigSection = "NLog" as input-parameter.



Sent from my Galaxy



From: Anatoliy Savinov ***@***.***>
Date: 27/06/2022 12:06 (GMT+01:00)
Subject: Re: [NLog/NLog.Web] Load config from json file with custom name (PR #805)


Something like this?
public static ISetupBuilder LoadConfigurationUsingConfigBuilder(this ISetupBuilder setupBuilder, Func<IConfigurationBuilder> builderFunc)

snakefoot:
The idea was to make `LoadConfigurationFromAppSettings` call the new method that takes `Func<IConfigurationBuilder>` as input-parameter. And then remove `LoadConfigurationFromJson`.

So `LoadConfigurationFromAppSettings` just calls the new method with this extracted into a local delegate:
```c#
            var builder = new ConfigurationBuilder()
                // Host Configuration
                .SetBasePath(currentBasePath)
                .AddEnvironmentVariables(prefix: "ASPNETCORE_")
                .AddEnvironmentVariables(prefix: "DOTNET_")
                // App Configuration
                .AddJsonFile("appsettings.json", optional, reloadOnChange)
                .AddJsonFile($"appsettings.{environment}.json", optional: true, reloadOnChange: reloadOnChange)
                .AddEnvironmentVariables();
```

Fluttercom:
Removed `LoadConfigurationFromJson` and still left the possibility to define custom filename.
Not sure about correct place of nlog.config fallback, as `LoadConfigurationUsingConfigBuilder` now requires two additional params.

snakefoot:
After looking at the code, then I think it is just easier if you do this in your own application:

```c#
var config = new ConfigurationBuilder().AddJsonFile("nlog.json").Build();
NLog.LogManager.Setup().SetupExtensions(e => e.RegisterNLogWeb()).LoadConfigurationFromSection(config);
```

PR #816, opened 2022-07-08
Add buffering target wrapper to ASP.NET Core

### LINKED ISSUE #304: Port AspNetBufferingWrapper to ASP.NET Core
From https://github.com/NLog/NLog/issues/2806

Port to middleware. 

Current docs: https://github.com/nlog/nlog/wiki/AspNetBufferingWrapper-target

bakgerman:
Attempt to move NLog.Web target wrapper to NLog.Web.AspNetCore, and resolve #304

Application must setup middleware `NLogBufferingTargetWrapperMiddleware`, which enables the use of target-wrapper `AspNetBufferingTargetWrapper` (`type="AspNetBufferingWrapper"`)


tests/NLog.Web.AspNetCore.Tests/AspNetCoreTests.cs
```diff
@@ -222,7 +222,7 @@ public void SkipRegisterHttpContext()
             Assert.Null(webhost.Services.GetService<IHttpContextAccessor>());
         }
 
-
+        /*
```
304NotModified, line 225: what happend here?

304NotModified (changes requested):
Could you please add some tests?

maybe helpful: https://stackoverflow.com/questions/59583406/how-to-unit-test-a-net-middleware-that-uses-response-onstarting

bakgerman:
> Could you please add some tests?
> 
> maybe helpful: https://stackoverflow.com/questions/59583406/how-to-unit-test-a-net-middleware-that-uses-response-onstarting

Certainly, thank you for the link.  I was looking for the existing unit tests for the ASP.NET version of the buffer target wrapper but could not locate.

tests/NLog.Web.AspNetCore.Tests/AspNetCoreTests.cs
bakgerman: Here something strange happened.  AppVeyor had an error message that the class MemoryTarget could not be found.  But I did not modify AspnetCoreTests as part of my PR.  I commented that method that uses the MemoryTarget so that AppVeyor would build.  I will revert that commit.

bakgerman: I see what happened, I added a Target namespace to the assembly, which confused the reference Target.MemoryTarget.  I changed it to NLog.Target.MemoryTarget and it builds again.  I will work on the unit tests now.

bakgerman:
Will need to determine why Write method in the target sees a null HttpContext if this is due to unit test only or if this fails in actual application.  Will work on that next.  Rest of it seems to be OK.

bakgerman:
I have made a commit so that the http context is saved into an httpcontextaccessor upon begin request, so that the http context is available during the write method.  Since we don't have httpcontext.current in asp.net core.  Debugging in the unit tests this did write events to the log buffer now.

snakefoot:
Curious how this works together with AsyncTargetWrapper. The NLog Layout provides automatic context-capture, but now you have to implement everything yourself.

But it also seems that the original NLog.Web-module is a little weird, since it explict calls `WrappedTarget.PrecalculateVolatileLayouts(logEvent.LogEvent);` which does not make any sense.

I can sadly enough not review this, since I don't have enough experience with neither NLog.Web or NLog.Web.AspNetCore. Not a big fan of static variables.

src/NLog.Web.AspNetCore/NLogBufferingMiddleware.cs
```diff
@@ -0,0 +1,65 @@
...
+namespace NLog.Web
+{
+    /// <summary>
+    /// This class is to intercept the HTTP pipeline and to allow additional logging of the following
+    ///
+    /// POST request body
+    ///
+    /// Usage: app.UseMiddleware&lt;NLogBufferingMiddleware&gt;(); where app is an IApplicationBuilder
+    /// </summary>
+    public class NLogBufferingMiddleware
```
snakefoot, line 15: Think the class-name is a little vague.

bakgerman:
> 

yes unfortunately any strange aspects of the original may exist in this one.  I was myself unable to understand how this will work in concert with the existing AsyncTargetWrapper.  This waits until the request is completed, but then executes all of the logging at the end, and what if there is a large number of logs buffered, how long can that take to execute?

src/NLog.Web.AspNetCore/NLogBufferingMiddleware.cs
bakgerman: Renamed to NLogBufferingTargetWrapperMiddleware

bakgerman:
Added a commit so that the asp.net and asp.net core version both use a common base class, so no more duplicated code.

bakgerman:
The code for ASP.NET and ASP.NET Core is now 95% the same in base classes.

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapperEventBase.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Event to be raised at the end of each HTTP Request.
+        /// </summary>
+        public static event EventHandler<HttpContextEventArgs> EndRequest;
+
+        /// <summary>
+        /// Invoke the begin request event handler
+        /// </summary>
+        /// <param name="eventArgs"></param>
+        public void InvokeBeginRequestHandler(HttpContextEventArgs eventArgs)
```
snakefoot, line 24: Can this become protected ?

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapperEventBase.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            BeginRequest?.Invoke(null, eventArgs);
+        }
+
+        /// <summary>
+        /// Invoke the end request event handler
+        /// </summary>
+        /// <param name="eventArgs"></param>
+        public void InvokeEndRequestHandler(HttpContextEventArgs eventArgs)
```
snakefoot, line 33: Can this become protected ?

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapperEventBase.cs
bakgerman: Complete

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapperEventBase.cs
bakgerman: Complete.  Also made the 2 events internal instead of public in same class

snakefoot:
@bakgerman Curious if you have ever used this code on ASP.NET Core Application? Or if you are just picking up a challenge?

I'm guessing the goal of this buffering-wrapper is to provide all LogEvents leading up to a certain state (probably error-state triggered by exception). And for all other situations then just discard the collected events?

bakgerman:
> @bakgerman Curious if you have ever used this code on ASP.NET Core Application? Or if you are just picking up a challenge?
> 
> I'm guessing the goal of this buffering-wrapper is to provide all LogEvents leading up to a certain state (probably error-state triggered by exception). And for all other situations then just discard the collected events?
> 
> I'm guessing there will be a lock-congestion/memory-allocation performance penalty when trying to enable this feature on ASP.NET Application that handles concurrent web-requests.

Yes, I am just picking up a challenge.  And try to make asp.net core have feature equality to asp.net.  

But this is indeed a strange target, that holds onto logs until an event occurs, then flushes them all at once.  

If we have many http requests at same time, and the log events are all held until request is over, then the result may not be good.  Logging should never affect the system being logged.

Wondering if we should mark AspNet BufferingTargetWrapper as Obsolete and not make AspNetCore version?  Does anyone know reason behind creation of the original class for Asp.net?

If this pull request is cancelled that is fine, I can examine request for IHealthCheck feature instead.

snakefoot:
Think it would make most sense if there was only one `AspNetBufferingTargetWrapper`, but depending on target-platform, then it hooked up to Http-Module or Http-Middleware. Maybe the Http-Module and Http-Middleware could implement an internal-interface, if it would make the common `AspNetBufferingTargetWrapper`-implementation-cleaner.

Would also prefer if it was possible to support ServiceFabric on AspNetCore, so one could have multiple web-sites running within the same host-application (Each with their own instance of NLog-Logging-Provider + NLog LogFactory-Instance + NLog Targets Instances)

bakgerman:
> Think it would make most sense if there was only one `AspNetBufferingTargetWrapper`, but depending on target-platform, then it hooked up to Http-Module or Http-Middleware. Maybe the Http-Module and Http-Middleware could implement an internal-interface, if it would make the common `AspNetBufferingTargetWrapper`-implementation-cleaner.
> 
> Would also prefer if it was possible to support ServiceFabric on AspNetCore, so one could have multiple web-sites running within the same host-application (Each with their own instance of NLog-Logging-Provider + NLog LogFactory-Instance + NLog Targets Instances). Maybe the HttpContext-usage should be different, so it is actually the target that initializes the buffer on first log-event-write (instead of middleware initializing the buffer upfront). But it is the middleware that flushes buffers on end-request.

Yes, I can certainly do the first part.  Then I will learn about ServiceFabric and think about the second part.

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
```diff
@@ -209,12 +209,37 @@ protected virtual void SaveHttpContext(HttpContextEventArgs httpContextEventArgs
 #endif
         }
 
+        private bool _firstWrite = true;
```
snakefoot, line 212: Every WebRequest will have their own HttpContext-scope, so not sure this makes any sense.

Was thinking `GetRequestBuffer` could be called `GetOrCreateBuffer`

bakgerman: I see, yes i will change that

bakgerman: Fixed as suggested

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
```diff
@@ -168,48 +202,65 @@ protected override void CloseTarget()
...
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="args"></param>
         private void OnEndRequest(object sender, EventArgs args)
         {
-            var buffer = GetRequestBuffer();
-            if (buffer != null)
+            var context = HttpContextAccessor.HttpContext;
```
snakefoot, line 246: Think this logic should be in the Middleware / HttpModule. So the it is the Middleware / HttpModule that performs the lookup in the active HttpContext for any active buffer and "commits" the buffered logevents.

The buffer should have an owner-reference to the `AspNetBufferingTargetWrapper`-target that created the buffer, so the Middleware / HttpModule can notify the owner that it is time to write.

Thus removing `EndRequest`-EventHandler.

snakefoot: But I guess that means the `DataSlot` must be static, and we should be able to handle a linked-list of buffers, when having multiple AspNetBufferingTargetWrapper-objects.

So the buffer should both have an "owner" but also a "next"-reference to allow having multiple `AspNetBufferingTargetWrapper`-target register multiple buffers for the same HttpContext-Item-Dataslot. And then the Middleware / HttpModule can notify the entire chain of buffers that HttpRequest has completed, and it is time to write.

bakgerman: We could do that certainly.  Maybe there is a way to keep DataSlot not static and still have the middleware/http module do the buffered log events.  
I see LoggingConfiguration has an AllTargets ReadOnlyCollection.  Could the middleware/ http module loop thru that collection and if the target is a AspNetBufferingTargetWrapper, call an internal method on the target that would commit the LogEvents in that target's own private DataSlot?  Could add an extension method to LoggingConfiguration called GetAllTargetsOfType<T>() for example as well.

bakgerman: Committed a change showing this approach.

snakefoot: I don't like global static variables, which is why I'm suggesting using the HttpContext as the transport for handling the request-buffering.

Replacing the static-event-handler with the global static LogManager is not a good trade. Also there is a performance penalty in LogFactory.Configuration-property, as it contains a monitor-lock.

bakgerman: Those are very valid points.  I have removed LogManager and instead use a Dictionary, this should be simialr to your request now.

bakgerman: The only static is an object used as a key to lookup the Dictionary in the HttpContext.Items.

src/NLog.Web.AspNetCore/NLogBufferingTargetWrapperMiddleware.cs
```diff
@@ -44,11 +44,12 @@ public async Task Invoke(HttpContext context)
             }
             finally
             {
-                var targets = LogManager.Configuration.AllTargets;
-                Parallel.ForEach(targets, target =>
```
snakefoot: Think the overhead from doing `Parallel.ForEach` is larger than just running on the actual thread.

Also by scheduling on a secondary thread, means that HttpContext can become disposed/unavailable, so it is impossible for the secondary thread to access the HttpContext-Items-dictionary (when it is actual executed).

bakgerman: Sure, we can just use normal for each.
Should we take all the code out of middleware and http module and just have static AspNetBufferWrapper.FlushAll method?

snakefoot:
Yes a static flush method on AspNetBufferWrapper would be good.



Sent from my Galaxy



Date: 26/07/2022 19:02 (GMT+01:00)
Subject: Re: [NLog/NLog.Web] Add buffering target wrapper to ASP.NET Core (PR #816)


@bakgerman commented on this pull request.

In src/NLog.Web.AspNetCore/NLogBufferingTargetWrapperMiddleware.cs<https://github.com/NLog/NLog.Web/pull/816#discussion_r930215873>:

> @@ -44,11 +44,12 @@ public async Task Invoke(HttpContext context)
             }
             finally
             {
-                var targets = LogManager.Configuration.AllTargets;
-                Parallel.ForEach(targets, target =>


Sure, we can just use normal for each.
Should we take all the code out of middleware and http module and just have static AspNetBufferWrapper.FlushAll method?

src/NLog.Web.AspNetCore/NLogBufferingTargetWrapperMiddleware.cs
snakefoot: Yes a static flush method on AspNetBufferWrapper would be good (Taking the HttpContext as input-parameter)

bakgerman: Completed

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
```diff
@@ -120,46 +135,119 @@ public int BufferGrowLimit
...
+
+        private IHttpContextAccessor RetrieveHttpContextAccessor(
+            IServiceProvider serviceProvider,
+            LoggingConfiguration loggingConfiguration)
         {
-            base.InitializeTarget();
+#if !ASP_NET_CORE
+            return new DefaultHttpContextAccessor();
+#else
+            return ServiceLocator.ResolveService<IHttpContextAccessor>(serviceProvider, loggingConfiguration);
```
snakefoot, line 171: Can the input parameters be removed so it just becomes:
```c#
return ServiceLocator.ResolveService<IHttpContextAccessor>(ResolveService<IServiceProvider>(), LoggingConfiguration);
```

bakgerman: Completed.  Also removed many #If ASP_NET_CORE statements and removed [NLogConfigurationIgnoreProperty] on HttpContextAccessor since this is internal and not public.

bakgerman: Does this HttpContextAccessor improvement also need to be performed on AspNetLayoutRenderBase?

snakefoot: No it is okay to leave `AspNetLayoutRenderBase` alone.

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// <summary>
+        /// The Key for the buffer dictionary in the HttpContext.Items collection
+        /// </summary>
+        private static readonly object HttpContextItemsKey = new object();
+
+        /// <summary>
+        /// The lock obtained before the dictionary is created or the slot
+        /// in the dictionary is created.
+        /// </summary>
+        private readonly object _lock = new object();
```
bakgerman, line 178: Should this be static?  What if the NLog.Config has several BufferingTargetWrappers?

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
```diff
@@ -0,0 +1,301 @@
...
+            {
+                return null;
+            }
+
+            var bufferDictionary = GetBufferDictionary(context);
+
+            // If the dictionary is missing, create that first
+            if (bufferDictionary == null)
+            {
+                lock (_lock)
```
bakgerman, line 200: Should we just say lock(context)?

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
```diff
@@ -0,0 +1,301 @@
...
+                }
+            }
+
+            // if the slot for this class instance is missing, create that first
+            if (bufferDictionary.ContainsKey(this))
+            {
+                return bufferDictionary[this];
+            }
+
+            lock (_lock)
```
bakgerman, line 212: Should we just say lock(context)?

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
snakefoot: In the "simple" world, then one could could get rid of the dictionary, and just have each target-wrapper to have their own key. In the "real" world, then HttpContext-Items is actually not threadsafe, and the correct thing would be that the middleware upfront created a dictionary and inserted into the HttpContext-Items (using a static key). And then target-wrappers would use the same static-key to insert themselves in the already created dictionary.

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
snakefoot: See comment above about always creating empty dictionary upfront in middleware, so target-wrapper will only insert itself into the pre-created dictionary on logevent.

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
snakefoot: See comments above.

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
bakgerman: Ok that makes sense, I will modify that.

bakgerman: Completed, HttpModule and Middleware calls Buffering.TargetWrapper.Initialize(HttpContext) at startup phase.

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
bakgerman: Completed, see above.

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
bakgerman: Completed, see above.

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
snakefoot: Think `_lock` shoudl be removed, and instead once should perform the synchronization around the pre-allocated dictionary.

bakgerman: Completed, did not catch that as a remnant from the previous method used.

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
```diff
@@ -0,0 +1,311 @@
...
+            lock (bufferDictionary)
+            {
+                if (!bufferDictionary.ContainsKey(this))
+                {
+                    bufferDictionary.Add(this,
+                        new Internal.LogEventInfoBuffer(BufferSize, GrowBufferAsNeeded, BufferGrowLimit));
+                }
+            }
+
+            return bufferDictionary[this];
```
snakefoot, line 233: Can this be moved inside the `lock`. Maybe use `bufferDictionary.TryGetValue` instead of `ContainsKey`

bakgerman: Done as requested

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
```diff
@@ -223,14 +223,14 @@ private Internal.LogEventInfoBuffer GetOrCreateRequestBuffer(HttpContextBase con
...
             lock (bufferDictionary)
             {
-                if (!bufferDictionary.ContainsKey(this))
+                Internal.LogEventInfoBuffer buffer;
+                if (!bufferDictionary.TryGetValue(this, out buffer))
                 {
                     bufferDictionary.Add(this,
                         new Internal.LogEventInfoBuffer(BufferSize, GrowBufferAsNeeded, BufferGrowLimit));
                 }
+                return bufferDictionary[this];
```
snakefoot, line 232: maybe do this instead (single lookup):
```c#
                if (!bufferDictionary.TryGetValue(this, out buffer))
                {
                    buffer = new Internal.LogEventInfoBuffer(BufferSize, GrowBufferAsNeeded, BufferGrowLimit)
                    bufferDictionary.Add(this, buffer);
                }
                return buffer;
```

bakgerman: Completed

snakefoot: Please re-add the `lock (bufferDictionary)` around this logic

bakgerman: Sorry about that.

snakefoot: No worries. Now just need to find some volunteers, that find it useful, and wants to test it.

bakgerman:
Should we merge this in, or close and leave unmerged?  Have we received any community feedback?

snakefoot:
> Should we merge this in, or close and leave unmerged? Have we received any community feedback?

I have not heard anything from anyone, even though we been poking those who once showed interest.

But I like the change and think it is an improvement of the existing logic. Could you "repair" the pull-request so it becomes a non-breaking change:
- Keep the old name `NLogHttpModule` (And not `NLogBufferingTargetWrapperModule`)
- Perform mixed reset of your pull-request so it becomes a single commit, and see if we can get `AspNetBufferingTargetWrapper.cs` to become a "move" instead of "delete" + "add".

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
```diff
@@ -0,0 +1,305 @@
...
+                WrappedTarget?.WriteAsyncLogEvent(logEvent);
+            }
+        }
+
+        /// <summary>
+        /// Called from the HttpModule or Middleware upon the end of the HTTP pipeline
+        /// Flushes all instances of this class registered in the HttpContext
+        /// </summary>
+        /// <param name="context"></param>
+        internal static void Flush(HttpContextBase context)
```
snakefoot, line 278: Could this be renamed to `RequestCompleted` ?

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
```diff
@@ -0,0 +1,305 @@
...
+            return new DefaultHttpContextAccessor();
+#endif
+        }
+
+        /// <summary>
+        /// Must be called by the HttpModule/Middleware upon starting.
+        /// This creates the dictionary in the HttpContext.Items with the proper key.
+        /// </summary>
+        /// <param name="context"></param>
+        internal static void Initialize(HttpContextBase context)
```
snakefoot, line 174: Could this be renamed to `BeginRequest` ?

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
bakgerman: Completed

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
bakgerman: Completed

bakgerman:
> > Should we merge this in, or close and leave unmerged? Have we received any community feedback?
> 
> I have not heard anything from anyone, even though we been poking those who once showed interest.
> 
> But I like the change and think it is an improvement of the existing logic. Could you "repair" the pull-request so it becomes a non-breaking change:
> 
> * Keep the old name `NLogHttpModule` (And not `NLogBufferingTargetWrapperModule`)
> * Perform mixed reset of your pull-request so it becomes a single commit, and see if we can get `AspNetBufferingTargetWrapper.cs` to become a "move" instead of "delete" + "add".

Kept the old name.  
Working on he git reset

bakgerman:
Moved to another pull request

PR #850, opened 2022-08-10
Improved re-entrancy Scope Lock for Session Value Layout Render

### LINKED ISSUE #847: AspNetSessionValueLayoutRenderer should be threadsafe
AspNetSessionValueLayoutRenderer modifies the HttpContext.Items-collection, but this collection is not threadsafe.

Implement other threadsafe way of doing re-entrant checking. Ex. like this: https://github.com/getsentry/sentry-dotnet/commit/e0d6b1de14e66406f0f62b86a1c6091cbb1c6349 (Only possible for ASP_NET_CORE or NET46)

bakgerman:
Also fix spelling errors and remove unused using statements. Resolves  #847


snakefoot:
Could the Re-Entrant logic be put in an IDiposable-struct, that implements IDisposable. Where struct-constructor setup the blocker, and the dispose-method removes the blocker.

Then the IDiposable-struct can used in a `using`-statement to protect the lookup.

src/Shared/Internal/RendererReEntrantManager.cs
```diff
@@ -0,0 +1,120 @@
...
+namespace NLog.Web.Internal
+{
+    /// <summary>
+    /// Manages if a LayoutRenderer can be called recursively.
+    /// Especially used by AspNetSessionValueLayoutRenderer
+    /// 
+    /// Since NET 35 does not support AsyncLocal or even ThreadLocal
+    /// a different technique must be used for that platform
+    /// </summary>
+    internal class RendererReEntrantManager : IDisposable
```
snakefoot, line 17: Can you change `internal class` to `internal struct` ? And remove `protected virtual void Dispose(bool disposing)`, so `void Dispose()` calls `DisposalImpl()` ?

bakgerman: ok

src/Shared/Internal/RendererReEntrantManager.cs
```diff
@@ -0,0 +1,120 @@
...
+            ReEntrantLock.Value = false;
+        }
+#endif
+
+        // Need to track if we were successful in the entry
+        // If we were not, we should not unlock in the dispose code
+        private bool entrySuccess;
+
+
+        internal bool TryGetLock()
```
snakefoot, line 73: Could `TryGetLock()` become private and called by the constructor, and then make new property `bool IsLockAcquired => entrySuccess` that is used instead from outside code.

src/Shared/Internal/RendererReEntrantManager.cs
bakgerman: changes uploaded

src/Shared/Internal/RendererReEntrantManager.cs
bakgerman: Done as requested

src/Shared/Internal/RendererReEntrantManager.cs
```diff
@@ -68,19 +71,17 @@ private void Unlock()
         private bool _obtainedLock;
```
snakefoot, line 71: Can this become `private readonly bool _obtainedLock` and assigned in the constructor ? (Return value from `TryGetLock()`)

src/Shared/LayoutRenderers/AspNetSessionValueLayoutRenderer.cs
```diff
@@ -93,36 +89,33 @@ protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
...
-#endif
-            if (value != null)
+
+            using (var reEntry = new RendererReEntrantManager(context))
             {
-                var formatProvider = GetFormatProvider(logEvent, Culture);
-                builder.AppendFormattedValue(value, Format, formatProvider, ValueFormatter);
+                if (!reEntry.IsLockAcquired)
+                {
+                    InternalLogger.Error(
```
snakefoot, line 107: Can this become `InternalLogger.Debug` and without using string-interpolation ? (Skip logging `logEvent.Message`)

src/Shared/Internal/RendererReEntrantManager.cs
bakgerman: Complete.

src/Shared/LayoutRenderers/AspNetSessionValueLayoutRenderer.cs
bakgerman: Complete

src/Shared/LayoutRenderers/AspNetSessionValueLayoutRenderer.cs
```diff
@@ -104,8 +104,8 @@ protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
             {
                 if (!reEntry.IsLockAcquired)
                 {
-                    InternalLogger.Error(
-                        $"Reentrant log event detected. Logging when inside the scope of another log event can cause a StackOverflowException. LogEventInfo.Message:{logEvent.Message}");
+                    InternalLogger.Debug(
+                        $"Reentrant log event detected. Logging when inside the scope of another log event can cause a StackOverflowException.");
```
snakefoot, line 108: Please also remove `$"` since not string-interpolation.

src/Shared/LayoutRenderers/AspNetSessionValueLayoutRenderer.cs
```diff
@@ -93,36 +89,33 @@ protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
...
             var context = HttpContextAccessor.HttpContext;
-            var contextSession = context.TryGetSession();
-            if (contextSession == null)
-                return;
-
-#if !ASP_NET_CORE
-            var value = PropertyReader.GetValue(item, contextSession, (session,key) => session.Count > 0 ? session[key] : null, EvaluateAsNestedProperties);
-#else
-            //because session.get / session.getstring also creating log messages in some cases, this could lead to stackoverflow issues. 
-            //We remember on the context.Items that we are looking up a session value so we prevent stackoverflows
```
snakefoot, line 104: Can the above comment be restored? So it appears just above `new RendererReEntrantManager` ?

src/Shared/LayoutRenderers/AspNetSessionValueLayoutRenderer.cs
bakgerman: done

src/Shared/LayoutRenderers/AspNetSessionValueLayoutRenderer.cs
bakgerman: done

src/Shared/Internal/RendererReEntrantManager.cs
```diff
@@ -0,0 +1,103 @@
...
+namespace NLog.Web.Internal
+{
+    /// <summary>
+    /// Manages if a LayoutRenderer can be called recursively.
+    /// Especially used by AspNetSessionValueLayoutRenderer
+    /// 
+    /// Since NET 35 does not support AsyncLocal or even ThreadLocal
+    /// a different technique must be used for that platform
+    /// </summary>
+    internal readonly struct RendererReEntrantManager : IDisposable
```
snakefoot, line 20: Can the class be renamed to ReEntrantScopeLock ?

bakgerman: Yes, that is more meaningful

src/Shared/Internal/ReEntrantScopeLock.cs
```diff
@@ -0,0 +1,103 @@
...
+            ReEntrantLock.Value = true;
+        }
+
+        private void Unlock()
+        {
+            ReEntrantLock.Value = false;
+        }
+#endif
+
+        private void DisposalImpl()
```
snakefoot, line 89: Can this be removed, and just have its logic in `void Dispose()` ?

src/Shared/Internal/ReEntrantScopeLock.cs
```diff
@@ -0,0 +1,103 @@
...
+    /// </summary>
+    internal readonly struct ReEntrantScopeLock : IDisposable
+    {
+        internal ReEntrantScopeLock(HttpContextBase context)
+        {
+            _httpContext = context;
+
+            // The line below is required, because this is a struct, otherwise we get CS0188 compiler error
+            // which is 'The 'this' object cannot be used before all of its fields have been assigned.'
+            _obtainedLock = false;
```
snakefoot, line 28: Can this assignment be removed, along with the comment above?

src/Shared/Internal/ReEntrantScopeLock.cs
bakgerman: Complete

src/Shared/Internal/ReEntrantScopeLock.cs
bakgerman: working on it...

bakgerman: If we do that we get this error:
[https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0188?f1url=%3FappId%3Droslyn%26k%3Dk(CS0188)](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0188?f1url=%3FappId%3Droslyn%26k%3Dk(CS0188))

If we make it a class, then that error is eliminated.

snakefoot: Yes one must assign all fields of a struct in the constructor, but right now you are assigning `_obtainedLock` two times. I would like to have just one assignment (The one from `TryGetLock`)

bakgerman: Changed as requested. But, seems .NET does not consider "variable = Method()' as a valid assignment.  The build error in AppVeyor matches the error i see in the VS 2022 output window.

snakefoot:
Yes see that TryGetLock must be changed to static method that receives httpContext as input parameter. Can this be done?



Sent from my Galaxy



Date: 12/08/2022 16:11 (GMT+01:00)
Subject: Re: [NLog/NLog.Web] Re-Entrancy management for Session Value Layout Render. (PR #850)


@bakgerman commented on this pull request.

In src/Shared/Internal/ReEntrantScopeLock.cs<https://github.com/NLog/NLog.Web/pull/850#discussion_r944504668>:

> +    /// <summary>
+    /// Manages if a LayoutRenderer can be called recursively.
+    /// Especially used by AspNetSessionValueLayoutRenderer
+    ///
+    /// Since NET 35 does not support AsyncLocal or even ThreadLocal
+    /// a different technique must be used for that platform
+    /// </summary>
+    internal readonly struct ReEntrantScopeLock : IDisposable
+    {
+        internal ReEntrantScopeLock(HttpContextBase context)
+        {
+            _httpContext = context;
+
+            // The line below is required, because this is a struct, otherwise we get CS0188 compiler error
+            // which is 'The 'this' object cannot be used before all of its fields have been assigned.'
+            _obtainedLock = false;


Changed as requested. But, seems .NET does not consider "variable = Method()' as a valid assignment. The build error in AppVeyor matches the error i see in the VS 2022 output window.

bakgerman:
changed as requested.  I did not know that about struct, so I learned something new.

src/Shared/Internal/ReEntrantScopeLock.cs
```diff
@@ -0,0 +1,130 @@
...
+        private static bool TryGetLock(HttpContextBase context)
+        {
+            // If context is null leave
+            if (context == null)
+            {
+                return false;
+            }
+
+            // If already locked, return false
+            if (IsLocked())
```
snakefoot, line 93: Can the logic in `IsLocked()` and `Lock()` be merged into `TryGetLock` ? (So there is only `TryGetLock()` and `Unlock`)

src/Shared/Internal/ReEntrantScopeLock.cs
```diff
@@ -0,0 +1,130 @@
...
+        // Need to track if we were successful in the lock
+        // If we were not, we should not unlock in the dispose code
+        internal bool IsLockAcquired { get; }
+
+#if NET35
+        private static readonly object ReEntrantLock = new object();
+
+        private static bool TryGetLock(HttpContextBase context)
+        {
+            // If context is null leave
```
snakefoot, line 39: Can the logic in `IsLocked()` and `Lock()` be merged into `TryGetLock` ? (So there is only `TryGetLock()` and `Unlock`)

snakefoot:
@bakgerman I think it looks great now. Do you have any more changes that you want to make ?

bakgerman:
> @bakgerman I think it looks great now. Do you have any more changes that you want to make ?

No, I'm done with this pull if you agree

snakefoot:
Once again thank you for being patient with all my nit-picking.

PR #856, opened 2022-08-15
aspnet-request-url - change to property-enum instead of booleans

### LINKED ISSUE #855: aspnet-request-url - change to property-enum instead of booleans
It is rather annoying to use `aspnet-request-url` for splitting the url into parts.

It could be nice if one could do this `${aspnet-request-url:properties=port}`

Instead of having to do this: `${aspnet-request-url:IncludeScheme=false:IncludeHost=false:IncludePath=false:IncludePort=true}`

The Properties would have to be a Flags-Enum, and then change all the existing boolean-properties to modify the value of Property-Flags-Enum. The default value of the Properties-value should match the current default-boolean-configuration.

bakgerman:
Resolves #855

### DIFF
--- a/src/Shared/Enums/AspNetRequestUrlProperty.cs
+++ b/src/Shared/Enums/AspNetRequestUrlProperty.cs
@@ -38,7 +38,7 @@ namespace NLog.Web.Enums
         /// <summary>
         /// To specify whether to include / exclude the Query string.
         /// </summary>
-        QueryString = 16,
+        Query = 16,
 
         /// <summary>
         /// By default, log the scheme://host/path
--- a/src/Shared/Enums/AspNetRequestUrlProperty.cs
+++ b/src/Shared/Enums/AspNetRequestUrlProperty.cs
@@ -16,7 +16,7 @@ namespace NLog.Web.Enums
         None = 0,
 
         /// <summary>
-        /// To specify whether to exclude / include the scheme.
+        /// To specify whether to exclude / include the scheme.  Ex. 'http' or 'https'
         /// </summary>
         Scheme = 1,
 
--- a/src/Shared/Enums/AspNetRequestUrlProperty.cs
+++ b/src/Shared/Enums/AspNetRequestUrlProperty.cs
@@ -1,0 +1,48 @@
+using System;
+
+namespace NLog.Web.Enums
+{
+    /// <summary>
+    /// Controls which portions of the URL are logged
+    /// This is a Flags enumeration
+    /// </summary>
+    [Flags]
+    public enum AspNetRequestUrlProperty
+    {
+        /// <summary>
+        /// Microsoft recommends a Flags enum to have a None=0 value.
+        /// See https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=net-6.0
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// To specify whether to exclude / include the scheme.
+        /// </summary>
+        Scheme = 1,
+
+        /// <summary>
+        /// To specify whether to exclude / include the host.
+        /// </summary>
+        Host = 2,
+
+        /// <summary>
+        /// To specify whether to include / exclude the Port.
+        /// </summary>
+        Port = 4,
+
+        /// <summary>
+        /// To specify whether to exclude / include the url-path.
+        /// </summary>
+        Path = 8,
+
+        /// <summary>
+        /// To specify whether to include / exclude the Query string.
+        /// </summary>
+        QueryString = 16,
+
+        /// <summary>
+        /// By default, log the scheme://host/path
+        /// </summary>
+        Default = Scheme|Host|Path
+    }
+}
--- a/src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
+++ b/src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
@@ -42,17 +42,10 @@ namespace NLog.Web.LayoutRenderers
         [Obsolete("Please use the Properties flags enumeration instead")]
         public bool IncludeQueryString 
         {
-            get => (Properties & AspNetRequestUrlProperty.QueryString) == AspNetRequestUrlProperty.QueryString;
+            get => GetFlagEnum(AspNetRequestUrlProperty.QueryString);
             set
             {
-                if (value)
-                {
-                    Properties |= AspNetRequestUrlProperty.QueryString;
-                }
-                else
-                {
-                    Properties &= ~AspNetRequestUrlProperty.QueryString;
-                }
+                SetFlagEnum(AspNetRequestUrlProperty.QueryString, value);
             }
         }
 
@@ -62,17 +55,10 @@ namespace NLog.Web.LayoutRenderers
         [Obsolete("Please use the Properties flags enumeration instead")]
         public bool IncludePort
         {
-            get => (Properties & AspNetRequestUrlProperty.Port) == AspNetRequestUrlProperty.Port;
+            get => GetFlagEnum(AspNetRequestUrlProperty.Port);
             set
             {
-                if (value)
-                {
-                    Properties |= AspNetRequestUrlProperty.Port;
-                }
-                else
-                {
-                    Properties &= ~AspNetRequestUrlProperty.Port;
-                }
+                SetFlagEnum(AspNetRequestUrlProperty.Port, value);
             }
         }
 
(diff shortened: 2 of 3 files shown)


src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
```diff
@@ -19,42 +18,123 @@ namespace NLog.Web.LayoutRenderers
...
         /// To specify whether to include / exclude the Query string. Default is false.
         /// </summary>
-        public bool IncludeQueryString { get; set; }
+        [Obsolete("Please use the Properties flags enumeration instead")]
+        public bool IncludeQueryString 
+        {
+            get => (Properties & AspNetRequestUrlProperty.QueryString) == AspNetRequestUrlProperty.QueryString;
+            set
+            {
+                if (value)
```
snakefoot, line 48: I know this is obsolete code, but could the setter call a private method, that takes two parameters as input (boolean + property-enum). Then the same private method could be used for all the obsolete-boolean-settings.

src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
```diff
@@ -89,56 +169,74 @@ protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
...
         private void RenderUrl(HttpRequest httpRequest, StringBuilder builder)
         {
-            if (IncludeScheme && !string.IsNullOrWhiteSpace(httpRequest.Scheme))
+            if (Properties == AspNetRequestUrlProperty.None)
+            {
+                return;
+            }
+
+            // We can use Enum.HasFlag(Enum) in .NET 4.0 and higher
+            if ( Properties.HasFlag(AspNetRequestUrlProperty.Scheme) &&
```
snakefoot, line 220: Very happy that you used the old-style bit-wise-operation for `ASP_NET_CORE`. Can the same be done here also, instead of using `HasFlag`.

src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
bakgerman: I see yes

src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
bakgerman: Yes that is cool

src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
bakgerman: also did a helper for the getter, that reduced the duplicated code also.

src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
```diff
@@ -122,20 +94,40 @@ public bool IncludeScheme
...
+                Properties &= ~bit;
+            }
+        }
+
+        /// <summary>
+        /// helper to get the bit in the flags enum
+        /// </summary>
+        /// <param name="bit"></param>
+        /// <returns></returns>
+        private bool GetFlagEnum(AspNetRequestUrlProperty bit)
```
snakefoot, line 126: Can it be renamed to `HasFlagEnum` ?

src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
```diff
@@ -122,20 +94,40 @@ public bool IncludeScheme
...
+                SetFlagEnum(AspNetRequestUrlProperty.Path, value);
             }
         }
 
+        /// <summary>
+        /// Helper to set the bit in the flags enum properly
+        /// </summary>
+        /// <param name="bit"></param>
+        /// <param name="flag"></param>
+        private void SetFlagEnum(AspNetRequestUrlProperty bit, bool flag)
```
snakefoot, line 109: Can it be renamed to `SetPropertiesFlag` ?

src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
bakgerman: ok, we will edit in real time !

src/Shared/Enums/AspNetRequestUrlProperty.cs
```diff
@@ -0,0 +1,48 @@
...
+    public enum AspNetRequestUrlProperty
+    {
+        /// <summary>
+        /// Microsoft recommends a Flags enum to have a None=0 value.
+        /// See https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=net-6.0
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// To specify whether to exclude / include the scheme.
```
snakefoot, line 19: // To specify whether to exclude / include the scheme. Ex. 'http:' or 'https:'

src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
```diff
@@ -89,56 +161,72 @@ protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
 
         private void RenderUrl(HttpRequestBase httpRequest, StringBuilder builder)
         {
+            if (Properties == AspNetRequestUrlProperty.None)
```
snakefoot, line 164: Think this should be removed, since unnecessary optimization.

src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
```diff
@@ -89,56 +161,72 @@ protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
...
+            else if (GetFlagEnum(AspNetRequestUrlProperty.QueryString))
             {
                 builder.Append(url.Query);
             }
         }
 #else
         private void RenderUrl(HttpRequest httpRequest, StringBuilder builder)
         {
-            if (IncludeScheme && !string.IsNullOrWhiteSpace(httpRequest.Scheme))
+            if (Properties == AspNetRequestUrlProperty.None)
```
snakefoot, line 190: Think this should be removed, since unnecessary optimization.

src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
```diff
@@ -19,42 +18,100 @@ namespace NLog.Web.LayoutRenderers
...
-        public bool IncludePath { get; set; } = true;
+        [Obsolete("Please use the Properties flags enumeration instead")]
+        public bool IncludePath
+        {
+            get => HasPropertiesFlag(AspNetRequestUrlProperty.Path);
+            set => SetPropertiesFlag(AspNetRequestUrlProperty.Path, value);
+        }
+
+        /// <summary>
+        /// Helper to set the bit in the flags enum properly
```
snakefoot, line 90: Please avoid XML-docs for private methods. Prefer that method-name explains what they do, than using comments.

src/Shared/LayoutRenderers/AspNetRequestUrlRenderer.cs
```diff
@@ -19,42 +18,100 @@ namespace NLog.Web.LayoutRenderers
...
+                Properties |= bit;
+            }
+            else
+            {
+                Properties &= ~bit;
+            }
+        }
+
+        /// <summary>
+        /// helper to get the bit in the flags enum
```
snakefoot, line 107: Please avoid XML-docs for private methods. Prefer that method-name explains what they do, than using comments.

src/Shared/Enums/AspNetRequestUrlProperty.cs
snakefoot, line 19: /// To specify whether to exclude / include the scheme. Ex. 'http' or 'https'

snakefoot (approved):
Think it looks great now. Thank you again for this fine contribution.

src/Shared/Enums/AspNetRequestUrlProperty.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// To specify whether to exclude / include the url-path.
+        /// </summary>
+        Path = 8,
+
+        /// <summary>
+        /// To specify whether to include / exclude the Query string.
+        /// </summary>
+        QueryString = 16,
```
snakefoot, line 41: @bakgerman In ASP.NET Core this is just called "Query". What do you prefer?

bakgerman: We should use the new Core values, especially if we are creating a new enum.

bakgerman: Changed to 'Query'

snakefoot:
Updated wiki: https://github.com/NLog/NLog/wiki/AspNetRequest-Url-Layout-Renderer

PR #889, opened 2022-11-23
Apply ArgumentNullException.ThrowIfNull to NLog source-code

bakgerman:
I try to follow your example exactly as you say.

I had to make if statement for this attribute

#if NETCOREAPP3_1_OR_GREATER
            [CallerArgumentExpression("argument")]
#endif

since it is not in NET 461 or CORE 2.x, but other than that I did not find any difficulty.


snakefoot:
Would it be possible to apply the solution found here: https://github.com/NLog/NLog/pull/5068 ? (Explicit define `CallerArgumentExpressionAttribute` when not part of target-platform)

Would be nice to get rid of these lines (Maybe always use `Guard.ThrowIfNull`):
```
#if NET6_0_OR_GREATER
using ArgumentNullException = System.ArgumentNullException;
#else
using ArgumentNullException = NLog.Web.Internal.ArgumentNullException;
#endif
```

bakgerman:
Should now be as requested, thank you for the reference into NLog commit.

bakgerman:
Added unit tests for CallerArgumentExpressionAttribute class

snakefoot:
Could you make an unit-test that verifies that `UseNLog()` throws a valid `ArgumentNullException` with expected `ParamName`-property ?

bakgerman:
unit test added as requested

snakefoot:
Looks like the internal `CallerArgumentExpressionAttribute` is working, since .NET Framework v4.6.1 also works.

Thank you this nice pull-request and the unit-tests.

PR #917, opened 2022-12-30
AspNetQueryStringLayoutRenderer - Added Exclude option

snakefoot:
Consistent with the other LayoutRenders that inherits from AspNetLayoutMultiValueRendererBase (And can print all items)


snakefoot:
Updated wiki - https://github.com/NLog/NLog/wiki/AspNetRequest-QueryString-Layout-Renderer

PR #921, opened 2023-01-18
Enable http status code layout renderer to output both the int and En…

### LINKED ISSUE #920: Extend ${aspnet-response-statuscode} to include Enum-ToString value
Introduce a Format option, when using Format="d" then it will output integer-value (Default)

Else cast to `HttpStatusCode` and output the [Enum.ToString(Format)](https://learn.microsoft.com/en-us/dotnet/api/system.enum.tostring). Ex. `304` becomes `NotModified`

https://github.com/NLog/NLog/wiki/AspNetResponse-StatusCode-Layout-Renderer

bakgerman:
…um.ToString() versions of the status code, using a Format property that defaults to integer. Resolves #920

### DIFF
--- a/src/Shared/LayoutRenderers/AspNetResponseStatusCodeRenderer.cs
+++ b/src/Shared/LayoutRenderers/AspNetResponseStatusCodeRenderer.cs
@@ -1,4 +1,6 @@
-﻿using System.Text;
+﻿using System;
+using System.Net;
+using System.Text;
 using NLog.Config;
 using NLog.LayoutRenderers;
 using NLog.Web.Internal;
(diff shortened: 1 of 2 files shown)


src/Shared/LayoutRenderers/AspNetResponseStatusCodeRenderer.cs
```diff
@@ -24,12 +38,23 @@ protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
...
-            if (statusCode >= 100 && statusCode <= 599)
+            // .NET format strings are case-insensitive
+            // See https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings
+            if (string.Equals(Format, IntegerFormat, StringComparison.InvariantCultureIgnoreCase))
             {
                 builder.Append(statusCode);
             }
+            else
+            {
+                builder.Append(((HttpStatusCode)statusCode).ToString());
```
snakefoot, line 56: Can we use `ToString(Format)` so the following Format-values are supported: `"G" or "g", "D" or "d", "X" or "x", and "F" or "f"`. See also: https://learn.microsoft.com/en-us/dotnet/standard/base-types/enumeration-format-strings

bakgerman: This should be complete, I have added unit tests for each format value.

src/Shared/LayoutRenderers/AspNetResponseStatusCodeRenderer.cs
```diff
@@ -9,12 +11,25 @@ namespace NLog.Web.LayoutRenderers
     /// ASP.NET Response Status Code.
     /// </summary>
     /// <remarks>
-    /// <code>${aspnet-response-statuscode}</code>
+    /// <code>${aspnet-response-statuscode} emits the http status code as an int</code>
+    /// <code>${aspnet-response-statuscode:Format=value} where value is a valid enumeration format string emits the specified format</code>
```
snakefoot, line 15: Can the docs about the possible format-values be moved down as `<remark>` for the `Format`-option?

src/Shared/LayoutRenderers/AspNetResponseStatusCodeRenderer.cs
snakefoot: Thank you for the extra testing of the Enum-Format-option.

src/Shared/LayoutRenderers/AspNetResponseStatusCodeRenderer.cs
bakgerman: good suggestion, completed

snakefoot: Thank you. It looks good.

snakefoot (approved):
I think it looks good. Do you have any  more changes, before merge ?

bakgerman:
no, i am good with the merge

snakefoot:
Updated wiki - https://github.com/NLog/NLog/wiki/AspNetResponse-StatusCode-Layout-Renderer

PR #923, opened 2023-01-19
Handle malformed X-Forwarded-For Header

bakgerman:
At the firm I work for, the network engineers have not followed X-Forwarded-For HHTP headers standards.
Instead of having the desired client IP at the start of the list of IP addresses in the header, the client IP is the second or third element, depending upon whether you are in development, test, or production network.

One way to solve this is to add an int Index property to the layout renderer.

### DIFF
--- a/src/Shared/LayoutRenderers/AspNetRequestIpLayoutRenderer.cs
+++ b/src/Shared/LayoutRenderers/AspNetRequestIpLayoutRenderer.cs
@@ -17,9 +17,9 @@ namespace NLog.Web.LayoutRenderers
     /// <remarks>
     /// <code>${aspnet-request-ip}</code> to return the Remote IP
     /// <code>${aspnet-request-ip:CheckForwardedForHeader=true}</code> to return first element in the X-Forwarded-For header
-    /// <code>${aspnet-request-ip:CheckForwardedForHeader=true:Index=1}</code> to return second element in the X-Forwarded-For header
-    /// <code>${aspnet-request-ip:CheckForwardedForHeader=true:Index=-1}</code> to return last element in the X-Forwarded-For header
-    /// <code>${aspnet-request-ip:CheckForwardedForHeader=true:Index=1:ForwardedForHeader=myHeader}</code> to return second element in the myHeader header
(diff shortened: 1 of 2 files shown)


src/Shared/LayoutRenderers/AspNetRequestIpLayoutRenderer.cs
```diff
@@ -35,6 +37,15 @@ public class AspNetRequestIpLayoutRenderer : AspNetLayoutRendererBase
...
         public bool CheckForwardedForHeader { get; set; }
 
+        /// <summary>
+        /// Gets or sets the array index of the X-Forwarded-For header to use, if the desired client IP is not at
+        /// the zeroth index.  Defaults to zero.  If the index is too large the last array element is returned instead.
+        /// If a negative index is used, this is used as the position from the end of the array.
+        /// Minus one will indicate the last element in the array.  If the negative index is too large the first index
+        /// of the array is returned instead.
+        /// </summary>
+        public int Index { get; set; } = 0;
```
snakefoot, line 47: With all the nice contributions that you have made, then I'm very eager to help you out.

But I think `Index` is a little too generic for such an exotic option.

Could it be called `CheckForwardedForHeaderOffset` ? And be implemented like this (where activates offset +  header-lookup at the same time):

```c#
public int CheckForwardedForHeaderOffset
{
     get => _checkForwardedForHeaderOffset;
     set
     {
           _checkForwardedForHeaderOffset = value;
           CheckForwardedForHeader = true;
     }
}
```
And can be used like this:
```xml
${aspnet-request-ip:CheckForwardedForHeaderOffset=1}
```

bakgerman: great idea, i will do that.  And thank you very much for all the help you provide me.

bakgerman: Completed

snakefoot:
NLog.Web.AspNetCore v5.2.2 has been released - https://www.nuget.org/packages/NLog.Web.AspNetCore

PR #977, opened 2023-08-05
Issues 934 and 935 for diagnostics

bakgerman:


snakefoot:
I don't like the overhead of constantly inserting and checking the HttpContext-Items.

Maybe something more basic like assigning a static-property, when Middleware- / HttpModule-constructor had been called, and then check the static-property during initialize ?

PR #979, opened 2023-08-05
Issue 934 and 935, Middleware Initialization Debugging

### LINKED ISSUE #954: Better diagnostics for AspNetRequestPostedBodyLayoutRenderer without HttpModule or MiddleWare
When initializing AspNetRequestPostedBodyLayoutRenderer then it should output a warning when HttpModule or Middleware is not configured.

bakgerman:
Uses ctor for middleware and httpmodule, and initializelayoutrender/initializetarget, instead of warning every time, as requested.

Resolves #954 and resolves #935

### DIFF
--- a/src/Shared/LayoutRenderers/AspNetRequestPostedBodyLayoutRenderer.cs
+++ b/src/Shared/LayoutRenderers/AspNetRequestPostedBodyLayoutRenderer.cs
@@ -1,5 +1,6 @@
 ﻿using System.Text;
 using NLog.LayoutRenderers;
+using NLog.Common;
 #if ASP_NET_CORE
 using Microsoft.AspNetCore.Http;
 #else
@@ -17,10 +18,27 @@ namespace NLog.Web.LayoutRenderers
     [LayoutRenderer("aspnet-request-posted-body")]
     public class AspNetRequestPostedBodyLayoutRenderer : AspNetLayoutRendererBase
     {
+        internal static bool MiddlewareInstalled;
+
         /// <summary>
         /// The object for the key in HttpContext.Items for the POST request body
         /// </summary>
         internal static readonly object NLogPostedRequestBodyKey = new object();
+
+        /// <inheritdoc />
+        protected override void InitializeLayoutRenderer()
+        {
+            if (!MiddlewareInstalled)
+            {
+#if ASP_NET_CORE
+                InternalLogger.Warn("NLogRequestPostedBodyMiddleware must be installed to use aspnet-request-posted-body in a layout.");
+#else
+                InternalLogger.Warn("NLogRequestPostedBodyModule must be installed to use aspnet-request-posted-body in a layout.");
+#endif
+            }
+
+            base.InitializeLayoutRenderer();
+        }
 
         /// <inheritdoc/>
         protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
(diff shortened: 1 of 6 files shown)


snakefoot:
Maybe having the alert in the initialize-section might produce too many false-positives.

Maybe it should be the on the first LogEvent where HttpContext is available.

Also could it become `InternalLogger.Info` instead of `InternalLogger.Warn`, and something like `"NLogRequestPostedBodyMiddleware is not yet initialized, which is required by aspnet-request-posted-body."`.

bakgerman:
Changed to first log event

src/Shared/LayoutRenderers/AspNetRequestPostedBodyLayoutRenderer.cs
```diff
@@ -17,14 +18,44 @@ namespace NLog.Web.LayoutRenderers
...
+
+        private static bool LogEventProcessed;
+        private static readonly object LogEventProcessedLock = new object();
+
         /// <inheritdoc/>
         protected override void DoAppend(StringBuilder builder, LogEventInfo logEvent)
         {
+            if (!LogEventProcessed)
+            {
+                lock (LogEventProcessedLock)
```
snakefoot, line 42: Though I like the idea of only printing the alert exactly once, then I think using a monitor-lock-object is a little too heavy. Could it just be (Removing `LogEventProcessedLock`):
```c#
if (!LogEventProcessed)
{
   LogEventProcessed = true;
   if (!MiddlewareInstalled)
   {
       ...
   }
}
```

src/Shared/LayoutRenderers/AspNetRequestPostedBodyLayoutRenderer.cs
```diff
@@ -17,14 +18,44 @@ namespace NLog.Web.LayoutRenderers
...
         /// </summary>
         internal static readonly object NLogPostedRequestBodyKey = new object();
 
+        /// <inheritdoc />
+        protected override void InitializeLayoutRenderer()
+        {
+            base.InitializeLayoutRenderer();
+        }
+
+        private static bool LogEventProcessed;
```
snakefoot, line 34: Can this become non-static, and be re-named to `_verfiedMiddlewareInstalled`

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
```diff
@@ -193,12 +194,34 @@ protected override void CloseTarget()
             base.CloseTarget();
         }
 
+        private static bool LogEventProcessed;
```
snakefoot, line 197: Can this become non-static, and be re-named to `_verfiedMiddlewareInstalled`

src/Shared/Targets/Wrappers/AspNetBufferingTargetWrapper.cs
```diff
@@ -193,12 +194,34 @@ protected override void CloseTarget()
...
+
         /// <summary>
         /// Adds the specified log event to the buffer.
         /// </summary>
         /// <param name="logEvent">The log event.</param>
         protected override void WriteAsyncThreadSafe(AsyncLogEventInfo logEvent)
         {
+            if (!LogEventProcessed)
+            {
+                lock (LogEventProcessedLock)
```
snakefoot, line 208: Though I like the idea of only printing the alert exactly once, then I think using a monitor-lock-object is a little too heavy. Could it just be (Removing `LogEventProcessedLock`):
```c#
if (!LogEventProcessed)
{
   LogEventProcessed = true;
   if (!MiddlewareInstalled)
   {
       ...
   }
}
```

snakefoot (commented):
Think it looks great. Just have some small suggestions.

bakgerman:
Thank you for the always good suggestions.  Modified as suggested.

(Have you pondered, when it may be possible to work on items marked as 'breaking changes' in the issues list?)

snakefoot:
Looks good. Thank you for another nice contribution.

PR #1096, opened 2025-06-26
NET Framework v4.6.2 with UseNLog for IHostApplicationBuilder

snakefoot:


snakefoot:
Even though Microsoft.AspNetCore v2.3.0 reference NET8-nuget-dependencies, then there is no `Microsoft.Extensions.Hosting.IHostEnvironment` when using `Microsoft.AspNetCore.WebHost.CreateDefaultBuilder()`

